PR #10, opened 2021-11-18
UDX-6051: add support for optional metal LB

christopherigm2:


k8_cortx_cloud/solution.yaml
```diff
@@ -21,6 +21,15 @@ solution:
     zookeeper: 3.7.0-debian-10-r182
     rancher: rancher/local-path-provisioner:v0.0.20
   common:
+    metallb:
```
osowski, line 24: We should add in an `enabled` field in here to use for the deployMetalLb function in deploy-cortx-cloud.sh , as that will allow us to leave the block of sensible defaults here but a default `enabled` value of `false`.

k8_cortx_cloud/solution.yaml
```diff
@@ -21,6 +21,15 @@ solution:
...
+      svc_name: cortx-metal-lb
+      name: default
+      protocol: layer2
+      addresses:
```
osowski, line 28: Per https://metallb.universe.tf/installation/#installation-with-helm, this should be a YAML list in the MetalLB values files. We do some conversion of that in the deploy-cortx-cloud logic, but are currently limited to 4 IP Addresses based upon the hard-coding here. Is there a specific reason why we didn't include this as an unbound list that would require less manual conversion into the MetalLb values format?

osowski (commented):
Added a few comments based on our previous conversations.

osowski:
Closing this PR for now. We will revisit additional networking support in the future.

PR #15, opened 2021-11-22
UDX-6642: Fix for Support Bundle tarfiles are not included in final t…

gargpalak048:
Include generated SB tarfiles in the final output tarball generated by log collector script

### DIFF
--- a/k8_cortx_cloud/logs-cortx-cloud.sh
+++ b/k8_cortx_cloud/logs-cortx-cloud.sh
@@ -58,6 +58,7 @@ function getInnerLogs()
   name="logs-${date}-${1}"
   logs_output=$(kubectl exec ${1} -- cortx_support_bundle generate -t file://${path} -b ${name} -m ${name})
   kubectl cp $1:$path/$name ./${logs_folder}
+  tar rf $logs_folder.tar ./${logs_folder}
   kubectl exec ${1} --namespace="${namespace}" -- bash -c "rm -rf ${path}"
 }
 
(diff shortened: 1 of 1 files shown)


k8_cortx_cloud/logs-cortx-cloud.sh
```diff
@@ -1,98 +1,99 @@
...
+  fi
+}
+
+function getInnerLogs()
+{
+  path="/var/cortx/support_bundle"
+  name="logs-${date}-${1}"
+  logs_output=$(kubectl exec ${1} -- cortx_support_bundle generate -t file://${path} -b ${name} -m ${name})
+  kubectl cp $1:$path/$name ./${logs_folder}
+  tar rf $logs_folder.tar ./${logs_folder}
```
gargpalak048, line 61: Added #L61 as a fix.
@cdeshmukh

cdeshmukh (changes requested):

k8_cortx_cloud/logs-cortx-cloud.sh
gargpalak048: ${logs_folder} will contain the SB tarfiles, we are adding it to the tar in a similar way saveLogs() & savePodDeatils() is adding to it.

Please refer to #L38 & #L50

gargpalak048: Also, changed the output file from tar to tar.gz compressed file

PR #30, opened 2021-12-08
Udx 6686 third party resource allocation

danthanhton:
Add the ability to modify resource limits and requests for the 3rd party deployment (cpu, memory, and storage size)


osowski (approved):
looks good to me!

PR #31, opened 2021-12-09
UDX-6831 - Add NodePort vs LoadBalancer flag support

osowski:
Signed-off-by: Rick Osowski <rosowski@gmail.com>

Ability for users to select LoadBalancer versus NodePort exposure for external facing services (`cortx-*-loadbal-svc-*`).

### DIFF
--- a/k8_cortx_cloud/solution.yaml
+++ b/k8_cortx_cloud/solution.yaml
@@ -39,6 +39,8 @@ solution:
       durability:
         sns: 1+0+0
         dix: 1+0+0
+    external_services:
+      type: LoadBalancer
   storage:
     cvg1:
       name: cvg-01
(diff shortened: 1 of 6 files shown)


k8_cortx_cloud/solution.yaml
```diff
@@ -39,6 +39,8 @@ solution:
       durability:
         sns: 1+0+0
         dix: 1+0+0
+    external_services:
+      type: LoadBalancer
```
danthanhton, line 43: Do you expect this is a required field? If it's a temporary field then it's not required to be added to the "solution-check.yaml" file.

osowski: It's something that we will want to expose at the highest level, even if the default is accepted. Right now, we don't have a way that simulates the functionality of the Helm values.yaml files that would allow us to have the user provide only the non-default values they want and everything else is defined in an internal values.yaml file in the chart. Since we are pulling all of our configuration information directly from the provided solution.yaml, I want to prevent building more logic into the bash scripts that will be going away in the future. The more logic we build into the solution.yaml and the associated charts, the better we'll be in the future.

danthanhton (approved):

PR #37, opened 2021-12-15
UDX-6949_improve_deploy_destroy_script_robustness

danthanhton:
- detect 'NotReady' worker nodes at the beginning of the deploy/destroy script and ask the users whether they want to proceed or exit the script early (UDX-6949)
- in deploy script, exit early if InitContainer fails to run (UDX-6921)

### DIFF
--- a/k8_cortx_cloud/deploy-cortx-cloud.sh
+++ b/k8_cortx_cloud/deploy-cortx-cloud.sh
@@ -32,29 +32,69 @@ max_consul_inst=3
 max_kafka_inst=3
 num_openldap_replicas=0 # Default the number of actual openldap instances
 num_worker_nodes=0
+not_ready_node_list=[]
+not_ready_node_count=0
 # Create a file consist of a list of node info and up to 'max_openldap_inst'
 # number of nodes. This file is used by OpenLDAP helm chart and will be deleted
 # at the end of this script.
 while IFS= read -r line; do
-    IFS=" " read -r -a node_name <<< "$line"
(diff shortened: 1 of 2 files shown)


k8_cortx_cloud/deploy-cortx-cloud.sh
```diff
@@ -32,30 +32,70 @@ max_consul_inst=3
...
-                num_openldap_replicas=$num_worker_nodes
-                node_list_info_path=$(pwd)/cortx-cloud-3rd-party-pkg/openldap/node-list-info.txt
-                if [[ -s $node_list_info_path ]]; then
-                    printf "\n" >> $node_list_info_path
-                fi
-                printf "$node_list_str" >> $node_list_info_path
+    IFS=" " read -r -a my_array <<< "$line"
+    node_name="${my_array[0]}"
+    node_status="${my_array[1]}"
+    if [[ "$node_name" == "NAME" && "$node_status" == "STATUS" ]]; then
```
osowski, line 44: This is fine as-is, but you can add the `--no-headers` flag to your `kubectl get` commands and that will prevent this from showing up in the response to be parsed.

osowski (approved):
LGTM! I added a comment for some future simplification of scripts to prevent header output when it's intended to be discarded, but that doesn't affect this implementation much right now.

k8_cortx_cloud/deploy-cortx-cloud.sh
danthanhton: cool, I didn't know that

PR #40, opened 2021-12-17
UDX-6813 implement cortx resource allocation

danthanhton:
Implement Cortx resource limits and requests. UDX-6813 branch is based off UDX-6667, therefore this branch can't be merged back into integration without merging UDX-6667 into integration first.

I have only tested the deployment with dummy containers. My VMs are not powerful enough to test CORTX containers. Also UDX-6667 changes are not finalized yet as I have encountered issues deploying CORTX containers using CORTX images.


osowski (commented):
Everything looks good. The only edit I would make is changing the root level construct to just be "resources" from "resource_allocation", so it aligns with Helm Chart best practices & common naming a little more closely.

For example. `.Values.*.resources.*.resources.requests.cpu` is a cleaner, more straight-forward path - both for implementor and consumer.

PR #41, opened 2021-12-22
Added the submodule dispatch event

venkuppu-chn:
Signed-off-by: Venkatesh K <venkatesh.k@seagate.com>


.github/workflows/dispatch_submodule_update.yml
```diff
@@ -0,0 +1,34 @@
...
+    branches: [ main ]
+jobs:
+  dispatch:
+    strategy:
+      matrix:
+        repo: ['Seagate/cortx']
+    runs-on: ubuntu-latest
+    steps:
+      - name: Repository Dispatch
+        uses: peter-evans/repository-dispatch@v1
```
osowski, line 30: @venkuppu-chn For security purposes, we need to be forking the third-party actions we are using to control overall exposure instead of using them inline directly.

osowski (changes requested):
This should be a universal change across all CORTX repos to fork third-party GitHub Actions and reference the forked versions instead of using third-party repository-hosted actions directly.

cc: @johnbent

.github/workflows/dispatch_submodule_update.yml
venkuppu-chn: @osowski Did you mean to say that we should fork the third-party action under seagate organization, publish in marketplace and use it. If we do so the line has changed to below. 

uses: seagate/repository-dispatch@v1

osowski: For the most part, yes. We do NOT need to publish it the Action in Marketplace. We simply need to fork the code to reside underneath the Seagate organization and reference it in the way you mention above. The Actions do not need to be in the marketplace to be consumable, as the `uses` reference is simply a GitHub repository lookup.

venkuppu-chn: @osowski As suggested, we have forked and used the fork action in the workflow. Kindly review and let us know.

osowski (approved):
lgtm

PR #45, opened 2022-01-07
UDX-7058_start_script_hangs_if_pod_fails_to_restart

danthanhton:
- exit wait loop if an error is detected while a pod is restarting


osowski (approved):
LGTM

PR #53, opened 2022-01-15
EOS-27502: Change from bash loop wait to "kubectl wait" for CORTX pods

walterlopatka:
Please review.

I don't love having hard-coded timeouts in the script.  For reference:

- CORTX Control:  timeouts 120s/60s, actual observed time 40-60s
- CORTX Data: timeouts 120s/60s, actual observed time ~80s
- CORTX Server: timeouts 120s/60s, actual observed time ~80s
- CORTX HA: timeouts 30s/30s, actual observed time ~0s
("actual observed time" is on lab VMs in a 5-node cluster)

### DIFF
--- a/k8_cortx_cloud/deploy-cortx-cloud.sh
+++ b/k8_cortx_cloud/deploy-cortx-cloud.sh
@@ -510,6 +510,8 @@ function deployKafka()
         --values $TMP_KAFKA_ENVVARS_YAML  \
         --wait
 
+    rm $TMP_KAFKA_ENVVARS_YAML
+
     printf "\n\n"
 }
 
@@ -888,15 +890,13 @@ function deployCortxSecrets()
 
 function silentKill()
 {
-    kill $1
-    wait $1 2> /dev/null
+    kill "$1"
+    wait "$1" 2> /dev/null
 }
 
 function waitForAllDeploymentsAvailable()
 {
-    INIT_DELAY=$1
-    shift
-    NEXT_DELAY=$1
+    TIMEOUT=$1
     shift
     DEPL_STR=$1
     shift
@@ -908,10 +908,10 @@ function waitForAllDeploymentsAvailable()
     
     # Initial wait
     FAIL=0
-    kubectl wait --for=condition=available --timeout=$INIT_DELAY $@
+    kubectl wait --for=condition=available --timeout="$TIMEOUT" $@
     if [ $? -ne 0 ]; then
         # Secondary wait
-        kubectl wait --for=condition=available --timeout=$NEXT_DELAY $@
+        kubectl wait --for=condition=available --timeout="$TIMEOUT" $@
         if [ $? -ne 0 ]; then
             # Still timed out.  This is a failure
             FAIL=1
@@ -920,7 +920,7 @@ function waitForAllDeploymentsAvailable()
 
     silentKill $DOTPID
     trap - 0
-    ELAPSED=$(($SECONDS - $START))
+    ELAPSED=$((SECONDS - START))
     echo
     if [ $FAIL -eq 0 ]; then
         echo "Deployment $DEPL_STR available after $ELAPSED seconds"
@@ -971,7 +971,7 @@ function deployCortxControl()
 
 
     printf "\nWait for CORTX Control to be ready"
-    waitForAllDeploymentsAvailable 120s 60s "CORTX Control" deployment/cortx-control
+    waitForAllDeploymentsAvailable 300s "CORTX Control" deployment/cortx-control
     if [ $? -ne 0 ]; then
         echo "Failed.  Exiting script."
         exit 1
@@ -1028,7 +1028,7 @@ function deployCortxData()
         node_name=${node_name_list[i]}
         deployments[${#deployments[@]}]="deployment/cortx-data-${node_name}"
     done
-    waitForAllDeploymentsAvailable 120s 60s "CORTX Data" ${deployments[*]}
+    waitForAllDeploymentsAvailable 300s "CORTX Data" ${deployments[@]}
     if [ $? -ne 0 ]; then
         echo "Failed.  Exiting script."
         exit 1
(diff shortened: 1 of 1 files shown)


k8_cortx_cloud/deploy-cortx-cloud.sh
```diff
@@ -883,11 +886,55 @@ function deployCortxSecrets()
...
+    shift
+
+    START=$SECONDS
+    (while true; do sleep 1; echo -n "."; done)&
+    DOTPID=$!
+    trap "silentKill $DOTPID" 0
+    
+    # Initial wait
+    FAIL=0
+    kubectl wait --for=condition=available --timeout=$INIT_DELAY $@
```
walterlopatka, line 911: @osowski and I discussed this "double-check" idea (if I understood correctly).  But now that I look at it, I'm not sure it buys us anything.  I would consider removing the "secondary wait" and just having a single, simple timeout.

doc/cortx-aws-k8s-installation.md
```diff
@@ -195,6 +195,33 @@ for ip in `aws ec2 describe-instances --filters Name=tag:Name,Values=$ClusterTag
 At this stage the Kubernetes cluster should be fully operational
 
 ## 3 Install CORTX 
+
```
walterlopatka, line 198: I'm not sure how this file got in here.  I didn't intend to make this change.

k8_cortx_cloud/deploy-cortx-cloud.sh
```diff
@@ -1325,4 +1320,4 @@ if [[ $num_motr_client -gt 0 ]]; then
     deployCortxClient
 fi
 deployCortxServices
-cleanup
```
walterlopatka, line 1328: How should I think about this?  Is this (unintentional) change a problem?

walterlopatka:
Update: I have tested with 9-node cluster as well.  CORTX Data takes 130-190 seconds; CORTX Server takes about 130-200 seconds (but once 283s!).  So starting CORTX Data/Server is proportional with how many nodes are in the cluster.  I'm not sure what to do about the timeout.  I could increase it.  Or we could make it configurable.  Or it could be set through a formula.

k8_cortx_cloud/deploy-cortx-cloud.sh
```diff
@@ -883,11 +886,55 @@ function deployCortxSecrets()
...
+    # Initial wait
+    FAIL=0
+    kubectl wait --for=condition=available --timeout=$INIT_DELAY $@
+    if [ $? -ne 0 ]; then
+        # Secondary wait
+        kubectl wait --for=condition=available --timeout=$NEXT_DELAY $@
+        if [ $? -ne 0 ]; then
+            # Still timed out.  This is a failure
+            FAIL=1
+            return
```
walterlopatka, line 918: This return should not be here.

doc/cortx-aws-k8s-installation.md
osowski: This is fine. It's part of a previous merge that I did accepting a PR. So it's more about git process than anything. We can review in team call.

k8_cortx_cloud/deploy-cortx-cloud.sh
```diff
@@ -510,8 +510,11 @@ function deployKafka()
         --values $TMP_KAFKA_ENVVARS_YAML  \
         --wait
 
-    rm $TMP_KAFKA_ENVVARS_YAML
```
osowski, line 513: We want this line in there so we don't leave leftover ephemeral installation files. Unless there's some specific reason, I would add it back in.

k8_cortx_cloud/deploy-cortx-cloud.sh
osowski: We need that cleanup step to be there, so you'll want to add it back in.

k8_cortx_cloud/deploy-cortx-cloud.sh
osowski: We want to be somewhat resilient in how we handle failures. Doing a simple second check helps in a "micorservices" oriented way... let's give it a reasonable timeout (calculated based on deployment size). If it fails, try it again with the same timeout (as it will either be successful in less than that timeout if our timeout formula is accurate or it will fail again because it's always going to fail). Then the deployment will succeed if it just needs more time, but it will also fail if it's always going to fail.

osowski:
> Update: I have tested with 9-node cluster as well. CORTX Data takes 130-190 seconds; CORTX Server takes about 130-200 seconds (but once 283s!). So starting CORTX Data/Server is proportional with how many nodes are in the cluster. I'm not sure what to do about the timeout. I could increase it. Or we could make it configurable. Or it could be set through a formula.

Added a handful of comments throughout. If you can make some of those updates and push a commit, I'll run some tests locally on my environment to see what it looks like. I like the idea of a calculation for the timeouts based on cluster size. It's fine for the timeout to be a very high watermark, as the wait will always exit before the timeout if things are up and running successfully. We just don't want the timeout to be too short.

osowski (changes requested):
Handful of comments and changes requested throughout. Overall, good structural evolution; we'll just need to fine tune some of the timeouts etc.

k8_cortx_cloud/deploy-cortx-cloud.sh
walterlopatka: Thanks!  I can't believe I deleted this!  Will replace.

k8_cortx_cloud/deploy-cortx-cloud.sh
walterlopatka: It wasn't clear from this diff, but the difference is that a newline was removed in my change at the end of the file (this is the last line in the file)

osowski (approved):
lgtm! I ran multiple tests locally with this PR code, both happy and unhappy paths, and everything worked as expected.

PR #56, opened 2022-01-17
DCO PR check workflow for K8s

mukul-seagate11:
Signed-off-by: Mukul Malhotra <mukul.malhotra@seagate.com>

DCO PR check workflow


.github/workflows/dco-check.yml
```diff
@@ -0,0 +1,23 @@
...
+#
+name: DCO Check
+
+on: [pull_request]
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: tisonkun/actions-dco@v1.0
```
osowski, line 23: @mukul-seagate11  For security purposes, we need to be forking the third-party actions we are using to control overall exposure instead of using them inline directly.

osowski (changes requested):
This should be a universal change across all CORTX repos to fork third-party GitHub Actions and reference the forked versions instead of using third-party repository-hosted actions directly.

cc: @johnbent

.github/workflows/dco-check.yml
mukul-seagate11: @osowski, its addressed & corrected

osowski (approved):
lgtm!

PR #57, opened 2022-01-20
Update README.txt

johnbent:
Small edit to avoid the use of non-inclusive language.


osowski (approved):
This is good for now. I have a pending PR that will remove this block of documentation entirely, so it's not a major issue going forward.

PR #58, opened 2022-01-22
EOS-27541: Add Github Action for shellcheck

keithpine:
Lint all bash scripts using [shellcheck](https://github.com/koalaman/shellcheck) for any push to main or integration branches and PRs.

Some extra strict options are enabled in .shellcheckrc (v0.8.0 required). All of the enabled checks cover some aspects of several Bash style guides, although they aren't all identical:

  - [bahamas10/bash-style-guide](https://github.com/bahamas10/bash-style-guide)
  - [BashGuide/Practices](http://mywiki.wooledge.org/BashGuide/Practices)
  - [Google Shell Style Guide](https://google.github.io/styleguide/shellguide.html)

For now, the Action is configured to always succeed because of existing failures.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,4 +1,4 @@
-# Copyright (c) 2021 Seagate Technology LLC and/or its Affiliates
+# Copyright (c) 2022 Seagate Technology LLC and/or its Affiliates
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as published
 # by the Free Software Foundation, either version 3 of the License, or
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,20 @@
+name: Lint
+
+on:
+  push:
+    branches:
+     - integration
+     - main
+  pull_request: ~
+  workflow_dispatch:
+
+jobs:
+  shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run Shellcheck
+        uses: ludeeus/action-shellcheck@1.1.0
+        # Temporarily pass until files are fixed
+        continue-on-error: true
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,3 +1,17 @@
+# Copyright (c) 2021 Seagate Technology LLC and/or its Affiliates
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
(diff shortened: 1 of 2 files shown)


walterlopatka:
LGTM (though I admit I don't have experience with this)

walterlopatka (approved):

johnbent:
@keithpine and @walterlopatka , I think this might be redundant since Codacy is already attached to this repo.

Here is a picture showing the Codacy badge attached to the front README of this repo.
image
If you click on the badge, it will bring you to the Codacy dashboard for this repo. 

You'll also notice that Codacy is already doing checks on each pull request:
image

And, finally, this page suggests to me that Codacy includes shellcheck:
https://docs.codacy.com/related-tools/codacy-plugin-tools/

johnbent:
Confirmed. I logged in to Codacy and looked at settings for the k8s repo and here is what I see:

image

keithpine:
@johnbent Thanks for that info. A couple of questions:

1. The version of Shellcheck being used is quite old, can it be upgraded? There are a couple of style checks that I've enabled that are not supported in the Codacy version.
2. Looks like the Codacy run is not flagging the same number of issues, the issue counts are significantly less than what the manual run sees. Can we enable all of the checks (and reduce if needed)? Do you know how the config file setting works? Will it select the `.shellcheckrc` file if it's in the repository, that way it could be managed w/o access to the Codacy admin panel.
3. How do we get Codacy to fail the PR check if it detects these issues in the future?

johnbent:
Plus @mukul-seagate11 . Mukul, can you help answer Keith's questions above about Codacy?

osowski (changes requested):
I'm good with this as a practice. I understand Codacy may run some of this, but I would prefer to control it with a finer-grained approach and then remove it when we have minimal (if any) shell scripts in the future.

The only change I would request is that we need to fork https://github.com/ludeeus/action-shellcheck in a github.com/Seagate repository and reference it in the `uses` path that way, instead of directly referencing the remote repository to run code in our pipeline unsupervised.

mukul-seagate11:
> I'm good with this as a practice. I understand Codacy may run some of this, but I would prefer to control it with a finer-grained approach and then remove it when we have minimal (if any) shell scripts in the future.
> 
> The only change I would request is that we need to fork https://github.com/ludeeus/action-shellcheck in a github.com/Seagate repository and reference it in the `uses` path that way, instead of directly referencing the remote repository to run code in our pipeline unsupervised.

Sure, let me address the same with. https://github.com/Seagate/action-shellcheck

mukul-seagate11:
> @johnbent Thanks for that info. A couple of questions:
> 
> 1. The version of Shellcheck being used is quite old, can it be upgraded? There are a couple of style checks that I've enabled that are not supported in the Codacy version.
> 2. Looks like the Codacy run is not flagging the same number of issues, the issue counts are significantly less than what the manual run sees. Can we enable all of the checks (and reduce if needed)? Do you know how the config file setting works? Will it select the `.shellcheckrc` file if it's in the repository, that way it could be managed w/o access to the Codacy admin panel.
> 3. How do we get Codacy to fail the PR check if it detects these issues in the future?

Unfortunately, at the moment Codacy does not support configuration files for Shellcheck but yes can be addressed via github actions for Shellcheck

mukul-seagate11:
Shellcheck workflow has been addressed via https://github.com/Seagate/cortx-k8s/pull/71

osowski (approved):
lgtm

PR #60, opened 2022-01-25
EOS-27575: Avoid capturing container logs as variables

keithpine:
Sometimes the log output from a container is so large that capturing it
in a variable and printing it will cause bash to crash. Instead of
capturing, just redirect the log output directly to the destination
file. This does have a side-effect of creating an empty log file if a
pod/container has no logs, where previously it would skip the file if
the logs were empty.

Other fixes:

 - Remove leading `./` from paths in the output tar file
 - Fix some Shellcheck complaints
 - Use long options for `tar` and `cortx_support_bundle` commands
   (script self-documentation)
 - Reduce some code duplication


osowski (approved):
lgtm.

good catch on the capturing logs to variables.

PR #71, opened 2022-02-01
Shellcheck-action workflow

mukul-seagate11:
Signed-off-by: Mukul Malhotra <mukul.malhotra@seagate.com>

Shellcheck-action workflow


osowski (changes requested):
Can you add `integration` to a branch for triggering, as well as target the `integration` branch as the merge target branch for the PR (instead of `main`)?

Otherwise, it looks good.

keithpine:
I already have an open PR #58, which handles checking pull requests and pushes to both branches, and temporarily disables errors until we clean up all the failures. I'll update that to use the action fork, and add the copyright notice to the workflow file.

PR #80, opened 2022-02-02
EOS-27546: Silence 'Defaulting container name' log collection messages

keithpine:
When `kubectl exec` or `kubectl cp` are used without a container name, a
warning is printed and the first container is used automatically. This
is fine but the warnings might confuse the status of the results so we
prefer to avoid them.

For support bundle generation, any of the pod's containers are a valid
selection, so use the first one listed.

Note that for kubectl v1.21+, the
`kubectl.kubernetes.io/default-container` annotation could be used to
select a default container, thus avoiding any special logic to pick a
container.

### DIFF
--- a/k8_cortx_cloud/logs-cortx-cloud.sh
+++ b/k8_cortx_cloud/logs-cortx-cloud.sh
@@ -85,14 +85,16 @@ function savePodDetail()
 function getInnerLogs()
 {
   local pod="$1"
+  local container_arg=()
+  [[ -n $2 ]] && container_arg+=(--container "$2")
   local path="/var/cortx/support_bundle"
   local name="bundle-logs-${pod}-${date}"
 
   printf "\n ⭐ Generating support-bundle logs for pod: %s\n" "${pod}"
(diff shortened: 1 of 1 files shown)


k8_cortx_cloud/logs-cortx-cloud.sh
```diff
@@ -85,14 +85,16 @@ function savePodDetail()
 function getInnerLogs()
 {
   local pod="$1"
+  local container_arg=()
+  [[ -n $2 ]] && container_arg+=(--container "$2")
```
keithpine, line 89: This is probably not really needed, there shouldn't be any pod w/o a container, but in case something strange happened this would just go with the previous behavior of not specifying a container. I can remove it if it seems pointless.

osowski (approved):
lgtm

k8_cortx_cloud/logs-cortx-cloud.sh
osowski: I'm good with this being added. I'd rather add it to explicitly catch an issue that would otherwise be thrown to the output and missed by the user.

walterlopatka (approved):

PR #86, opened 2022-02-05
Simplify function deployCortxSecrets()

walterlopatka:
Today I was working in function deployCortxSecrets(). I really don't understand what is the intent behind the "for" loop.  It looks almost as if it is looking for possible multiple blocks of secrets.  But solution.yaml really only defines on block: {name: cortx-secrets, content: [ list of secrets ]}.

Unless there is a compelling reason to keep this logic, I think it should be simplified.


osowski (approved):
lgtm so far.

the structure of this section would imply the ability to create multiple secrets, but that would need to be a list in solution.yaml and it's not set up that way. so i'm for simplifying now and revisiting when we have a greater need in the future.

walterlopatka:
I am closing this as I have opened a separate PR for this work (from a different branch, different commit message).

PR #94, opened 2022-02-08
EOS-27425: Distruptive container image upgrades

keithpine:
Add a utility script to upgrade the container images used in a deployed CORTX cluster. This assumes the solution is the `cortx-all` images which are set uniformally for all containers, e.g. `cortx-all:2.0.0-624-custom-ci`.

Based on #77, this script scales down all the Pods to 0 replicas, updates all of the container images for each of the CORTX Deployments, and scales the Pods back to 1 replica. Scaling up and down are handled by existing scripts. No other upgrade logic is performed besides setting a new image.

We prefer to use Helm to manage all of this and keep a history of changes, but attempts to do so were not successul, so for now this approach is sufficient.

Note that this PR also introduces some improvements to how the solution config file is loaded. In order to discourage the use of the checked-in `solution.yaml` file, a file will be chosen based on the following priority:

1. A file provided with the `-s` command line option, e.g. `./upgrade-cortx-cloud.sh -s local.yaml ...`.
2. A file provided in environment variable `CORTX_SOLUTION_CONFIG_FILE`, e.g. `CORTX_SOLUTION_CONFIG_FILE=local.yaml ./upgrade-cortx-cloud.sh ...`.
3. No file provided, so the default is `solution.yaml`, e.g. `./upgrade-cortx-cloud.sh ...`.

The plan would be to implement the above behavior in the other scripts as well for consistency.

### DIFF
--- a/k8_cortx_cloud/upgrade-cortx-cloud.sh
+++ b/k8_cortx_cloud/upgrade-cortx-cloud.sh
@@ -1,0 +1,168 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+trap 'handle_error "$?" "${BASH_COMMAND:-?}" "${FUNCNAME[0]:-main}(${BASH_SOURCE[0]:-?}:${LINENO:-?})"' ERR
+handle_error() {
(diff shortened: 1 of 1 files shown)


walterlopatka (approved):

osowski (approved):
lgtm

PR #98, opened 2022-02-11
EOS-28037: Removed Ldap from cortx-test provisioner framework

kailashyadav11:
Signed-off-by: Kailash Yadav <kailash.yadav@seagate.com>


kailashyadav11:
No required from prvsnr side, this change will be handled by services team

PR #100, opened 2022-02-11
EOS-27447 -  clusterip svc for hax

osowski:
Implements a new ClusterIP Service for HAX-enabled Data & Server Pods.

### DIFF
--- a/k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-data/templates/cortx-data-pod.yaml
+++ b/k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-data/templates/cortx-data-pod.yaml
@@ -2,7 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   labels:
-    app: {{ .Values.cortxdata.name }}
+    app: {{ quote .Values.cortxdata.name }}
   name: {{ .Values.cortxdata.name }}
   namespace: {{ .Values.namespace }}
 spec:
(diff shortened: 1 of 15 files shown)


k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-data/templates/cortx-data-pod.yaml
```diff
@@ -19,6 +19,7 @@ spec:
         cortx.io/machine-id: {{ .Values.cortxdata.machineid.value }}
         {{- end }}
         cortx.io/service-type: cortx-data
+        cortx.io/hax-enabled: "true"
```
walterlopatka, line 22: Just curious... is there some significance to `true` being in quotes?  I see that in the line above `cortx-data` does not have quotes around it.  (Maybe quotes are best practice, but you're not changing other lines?)

k8_cortx_cloud/deploy-cortx-cloud.sh
```diff
@@ -1208,14 +1215,6 @@ function deployCortxClient()
     printf "\n\n"
 }
 
-function deployCortxServices()
```
walterlopatka, line 1211: Do you need to remove the corresponding function deleteCortxServices() in destroy-cortx-cloud.sh?

walterlopatka (changes requested):
Overall looks good. As I ask in an inline comment, do you need to remove `function deleteCortxServices()` in `destroy-cortx.cloud.sh` (since you removed `function deployCortxServices()`)?

k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-data/templates/cortx-data-pod.yaml
osowski: The right-hand side of the Labels are strings. They should always be quoted, but k8s / Go make it easy to recognize when there's a String there and automatically convert it - but it can't do that with `true` or `false` since boolean types are a thing in Go. So you have to quote them when using "true" or "false" in Label values.

I'll go back and add quotes to the labels we have in here. We should make sure we are quoting all of our Labels as best practice as part of @keithpine work in the Labels issue.

keithpine (approved):
Won't be any problem for me to incorporate this into my ConfigMap changes. 👍

osowski:
Validated with Jenkins build via http://eos-jenkins.colo.seagate.com/job/Cortx-Kubernetes/job/setup-cortx-cluster-solution-input/352/

PR #101, opened 2022-02-14
EOS-28075 - cortx-client Helm implementation updates

osowski:
Signed-off-by: Rick Osowski <rosowski@gmail.com>

Based on discussion in EOS-28075 and updates in the `cortx-test` branch here, these changes should be all-encompassing for the cortx-client requirements:
- Adding cortx-client Helm deployments when specified in solution.yaml
- Updating ConfigMaps with appropriate cortx-client endpoints & metadata
- Updating deploy/undeploy scripts for additional elements being deployed/cleaned up.

### DIFF
--- a/k8_cortx_cloud/deploy-cortx-cloud.sh
+++ b/k8_cortx_cloud/deploy-cortx-cloud.sh
@@ -588,6 +588,15 @@ function deployCortxConfigMap()
     auto_gen_path="$cfgmap_path/auto-gen-cfgmap-$namespace"
     mkdir -p $auto_gen_path
 
+    motr_client_endpoints="[]"
+    if [[ $num_motr_client -gt 0 ]]; then
+        motr_client_endpoints="\n"
+        for i in "${!node_name_list[@]}"
+        do
+            motr_client_endpoints="$motr_client_endpoints"$'\n'"  - ""tcp://cortx-client-headless-svc-""${node_name_list[$i]}"":21001"
+        done
+    fi
+
     # Generate config files
     for i in "${!node_name_list[@]}"; do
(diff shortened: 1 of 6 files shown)


k8_cortx_cloud/deploy-cortx-cloud.sh
```diff
@@ -1166,13 +1213,16 @@ function deployCortxClient()
...
             --set cortxclient.sslcfgmap.mountpath="/etc/cortx/solution/ssl" \
+            --set cortxclient.machineid.value="$cortxclient_machineid" \
             --set cortxclient.localpathpvc.name="cortx-client-fs-local-pvc-$node_name" \
             --set cortxclient.localpathpvc.mountpath="$local_storage" \
             --set cortxclient.localpathpvc.requeststoragesize="1Gi" \
             --set namespace=$namespace \
             -n $namespace
     done
 
+exit 1
```
keithpine, line 1224: Leftover debug code?

osowski: D'oh! Good catch. I could have sworn I removed that and it wasn't causing problems in my testing.

keithpine (approved):
Looks good as far as I can tell. If you want to make Codacy happy, I'd suggest doing that, otherwise, I can fix it later when I address all the others.

osowski:
> Looks good as far as I can tell. If you want to make Codacy happy, I'd suggest doing that, otherwise, I can fix it later when I address all the others.

Since we know a lot of this will be changing soon, I'd rather try to make that happen in one swoop instead of piecing it out across all PRs since that will make our merges that much more difficult.

PR #102, opened 2022-02-14
CORTX-28699: Update `cortx_support_bundle` command for log collection

keithpine:
The `cortx_support_bundle generate` command was recently changed to require the `--cluster_conf_path` argument. Not specifying it results in a cryptic Python error. Previously we relied on the default being selected.

Update the container exec command to specify the argument as newly required.

This is confirmed to work using image `cortx-all:2.0.0-632-PI61.2`.


walterlopatka (approved):
lgtm

osowski (approved):
lgtm

PR #106, opened 2022-02-16
CORTX-27406: Manage CORTX ConfigMap via Helm

keithpine:
The "cortx-configmap" is now a complete implementation of a Helm Chart. This allows Helm to manage the creation and deletion of the config maps, and generate their contents based on template values.

This change by itself does not fix CORTX-27406, it lays down the framework for adding that functionality. This will be done in a future PR.


k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-configmap/templates/_cluster.tpl
```diff
@@ -0,0 +1,86 @@
...
+        services:
+          - io
+      - name: hare
+    {{- with .Values.clusterStorageVolumes }}
+    storage:
+    {{- range $key, $val := . }}
+    - name: {{ $key }}
+      type: {{ $val.type }}
+      devices:
+        metadata: {{- toYaml $val.metadataDevices | nindent 10 }}
```
keithpine, line 26: Minor difference here from before, this creates an array instead of a scalar. The former is the format described in the spec. It appears to work the same as before.

k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-configmap/templates/tls-configmap.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cortx-ssl-cert-cfgmap-{{ .Release.Namespace }}
```
keithpine, line 4: This should really be something like `{{ include "cortx-configmap.fullname" . }}-ssl`, but I didn't want to make even more changes than I already have.

k8_cortx_cloud/deploy-cortx-cloud.sh
```diff
@@ -1120,13 +958,11 @@ function deployCortxServer()
 
     printf "\nWait for CORTX Server to be ready"
     # Wait for all cortx-data deployments to be ready
-    declare -a deployments
+    local deployments=()
     for i in "${!node_selector_list[@]}"; do
-        node_name=${node_name_list[i]}
-        deployments[${#deployments[@]}]="deployment/cortx-server-${node_name}"
```
keithpine, line 1126: This is an "unbound" variable in bash, which was causing problems for me. `declare -a` makes an array, not an associative array (which is `declare -A`), and since this is just an array, I changed the usage to that.

k8_cortx_cloud/deploy-cortx-cloud.sh
```diff
@@ -895,10 +736,10 @@ function deployCortxSecrets()
     secrets="$(./parse_scripts/yaml_extract_block.sh "${solution_yaml}" "${yaml_content_path}" 2)"
 
     new_secret_gen_file="${secret_auto_gen_path}/${secret_fname}.yaml"
-    cp "${cfgmap_path}/templates/secret-template.yaml" "${new_secret_gen_file}"
+    cp "${cfgmap_path}/other/secret-template.yaml" "${new_secret_gen_file}"
```
keithpine, line 739: Note that `secret-template.yaml` is still the "old" template format, and so it can't reside in `templates/` or else Helm tries to evaluate it.

k8_cortx_cloud/deploy-cortx-cloud.sh
```diff
@@ -571,312 +572,152 @@ function deleteStaleAutoGenFolders()
...
+                --set "clusterStorageVolumes.${cvg_name}.dataDevices[${idx}]=${cvg_dev}"
+            )
+        done
+    done
+
+    helm install \
+        "cortx-cfgmap-${namespace}" \
+        cortx-cloud-helm-pkg/cortx-configmap \
+        --set fullnameOverride="cortx-cfgmap-${namespace}" \
+        "${helm_install_args[@]}"
```
keithpine, line 690: Once concern here is that the list of `--set` arguments may get very large (.e.g. hundreds of disks, etc.). There is a limit to the size of the command line arguments. On the systems I'm using, that's 2MB, it could vary.

osowski: We can investigate https://helm.sh/docs/helm/helm_install/#helm-install options and look at supplying the node/disk layouts via a dynamically generated values.yaml file or if using the `--set-file` flag works to get around the 2MB limit. Using a dynamically generated values.yaml file is still a valid improvement and implementation, as it passes all the information directly to Helm still and not pointers to other locally generated files (ie `helm get values {release}` will get you the entirety of the values we need to deploy the chart, not pointers to temporary files that were previously created during a deploy run).

osowski:
Looks good to me so far! 

http://eos-jenkins.colo.seagate.com/job/Cortx-Kubernetes/job/setup-cortx-cluster-solution-input/358/  🎉

k8_cortx_cloud/deploy-cortx-cloud.sh
keithpine: Agreed, YAML values files are a reasonable alternative and that was actually the approach I considered first. The downside is we need to generate them in bash. 😅

osowski (approved):
lgtm.

Ran two manual deployments and verified `hctl status` due to Jenkins jobs seeming to fail due to network issues.

Only ask is for @keithpine to create an issue to revisit the potential `--set` limitation and reimplement via using a `values.yaml` file. The issue does not need to be completed asap, just have it in our backlog.

keithpine:
Created CORTX-28848.

PR #119, opened 2022-02-24
Add CloudFormation template

trshaffer:
I wrote a script to do CORTX deployments using CloudFormation for Seagate/cortx#1381. It's basically follows the QSG (with a few tweaks for automated installation). I also wrote a short doc explaining how to use it. This should be a very easy way for new users to get up and running with CORTX+Kubernetes in AWS.

-----
[View rendered README.md](https://github.com/trshaffer/cortx-k8s/blob/cloudformation/README.md)
[View rendered doc/cloudformation.md](https://github.com/trshaffer/cortx-k8s/blob/cloudformation/doc/cloudformation.md)


README.md
```diff
@@ -95,6 +95,7 @@ For reference material, we have provided existing Kubernetes deployment models t
 
 1.  [Seagate Internal Jenkins Job](http://eos-jenkins.mero.colo.seagate.com/job/Cortx-kubernetes/job/setup-kubernetes-cluster/)
 2.  [CORTX on AWS and Kubernetes - Quick Install Guide](doc/cortx-aws-k8s-installation.md)
+    -  [AWS CloudFormation Template](doc/cloudformation.md) for this setup
```
johnbent, line 98: Instead of a separate document, would this be better as a set of edits to the original AWS/k8s guide?

johnbent (changes requested):
I'm wondering if this would be better as a separate document or as a set of edits to Gregory's original document? @gregnsk

README.md
trshaffer: Just moved it into the original guide

doc/cortx-aws-k8s-installation.md
```diff
@@ -457,6 +457,56 @@ chmod +x s3-benchmark
 | S3 | Multiple IPs (cortx-server-clusterip-svc pods) | tcp/443, tcp/80
 | IAM | Multiple IPs (cortx-server-clusterip-svc pods) | tcp/9443
 
+## 6 CloudFormation Guide
+
+The steps in this guide can be automated using AWS CloudFormation.
```
johnbent, line 462: So, I'm not familiar with this full guide. Did you add a new section starting at line 460 which is a simpler way to do everything that is between lines 0-460? If so, wouldn't it be better to delete the old instructions which do it the complex way and replace the entire contents of this doc with only the newer better way?
Or, are the full instructions still needed? If so, will the majority of readers prefer the old way or the new way? Whichever is the preference should be first. And at the top it should say, "If you prefer the blah blah way, jump to this section(LINK)"

But I suspect we need to just delete the old way and replace it with the new simpler way. No?

johnbent (changes requested):

doc/cortx-aws-k8s-installation.md
trshaffer: There's still useful information in the full guide on how to choose/customize the configuration, and explanation about what's happening, so I wouldn't want to simply delete the rest. The new way is certainly easier, so if you think the majority of readers will just use CloudFormation, I can greatly shorten the guide

johnbent:
Yes, I think most user will prefer the simpler way so I would suggest to trim this page as much as possible. @gregnsk , please let us know if you disagree. But otherwise @trshaffer please proceed accordingly. Thanks! Remzi always told me, "You cannot oversimplify something." Which might be ambiguous. But what I think he meant was, "Simpler is always better."

osowski (changes requested):
Added a handful of comments that will help to align with concurrent changes, as well as future plans. 

In addition to inline comments, in an attempt to not overcomplicate the root `k8_cortx_cloud` directory, can you move these files to a `k8_cortx_cloud/templates/cloudformation` subdirectory?

We're looking to condense how many disparate files are initially presented to the user, so being able to structure these changes hierarchically will help as we can use the Cloudformation work as a template for other types of automation templating.

johnbent (approved):
My requested changes have been made. LGTM!

PR #120, opened 2022-02-24
CORTX-27406: Allow customization of external service ports

keithpine:
The ports exposed by external services can now be customized using `solution.yaml`. This includes:

 - S3 http and https port
 - CORTX Control API port (CSM https)

Configuration of the container ports and application ports (within those containers) is not configurable.

Additional changes:

- The solution file (and Helm charts) allows optional assignment of NodePorts for each of the service ports (applicable to NodePort and LoadBalancer service types). It is now simple to expose nodePorts by default, for test or development purposes, directly from the solution configuration. 
- S3 and CORTX Control service types now default to `NodePort` instead of `LoadBalancer`.
- Removed the default `nodePort` value for the cortx-control service.
- Fixed a bug where `nodePort` was configured for the cortx-control service even if the type was `ClusterIP`; this is invalid and causes an error during deployment
- Fixed a bunch of shellcheck complaints in the deployment script.

**Breaking Changes**
- The format of `solution.yaml` has changed with respect to external services. S3 and Control are now separate maps, each configurable independently.
- The CORTX Control service was hardcoding a `nodePort` value of `31169`. If you were relying on this being set, it must be done manually in `solution.yaml` now.

### DIFF
--- a/k8_cortx_cloud/deploy-cortx-cloud.sh
+++ b/k8_cortx_cloud/deploy-cortx-cloud.sh
@@ -892,7 +892,7 @@ function deployCortxControl()
         --set cortxcontrol.image="${control_image}" \
         --set cortxcontrol.service.loadbal.name="cortx-control-loadbal-svc" \
         --set cortxcontrol.service.loadbal.type="${control_service_type}" \
-        --set cortxcontrol.service.loadbal.ports.api="${control_service_ports_https}" \
+        --set cortxcontrol.service.loadbal.ports.https="${control_service_ports_https}" \
         --set cortxcontrol.cfgmap.mountpath="/etc/cortx/solution" \
         --set cortxcontrol.cfgmap.name="cortx-cfgmap-${namespace}" \
         --set cortxcontrol.cfgmap.volmountname="config001" \
--- a/k8_cortx_cloud/deploy-cortx-cloud.sh
+++ b/k8_cortx_cloud/deploy-cortx-cloud.sh
@@ -102,6 +102,31 @@ function extractBlock()
     echo "$(./parse_scripts/yaml_extract_block.sh $solution_yaml $1)"
 }
 
+#######################################
+# Get a scalar value given a YAML path in a solution.yaml file.
+# Arguments:
+#   A YAML path to lookup, e.g. "solution.common.external_services.s3.type".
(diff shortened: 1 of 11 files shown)


k8_cortx_cloud/deploy-cortx-cloud.sh
```diff
@@ -829,35 +873,41 @@ function deployCortxControl()
...
+    [[ -n ${control_service_nodeports_https} ]] && optional_values+=(--set cortxcontrol.service.loadbal.nodePorts.https="${control_service_nodeports_https}")
+
+    helm install "cortx-control-${namespace}" cortx-cloud-helm-pkg/cortx-control \
         --set cortxcontrol.name="cortx-control" \
-        --set cortxcontrol.image=$cortxcontrol_image \
+        --set cortxcontrol.image="${control_image}" \
         --set cortxcontrol.service.loadbal.name="cortx-control-loadbal-svc" \
-        --set cortxcontrol.service.loadbal.type="$external_services_type" \
+        --set cortxcontrol.service.loadbal.type="${control_service_type}" \
+        --set cortxcontrol.service.loadbal.ports.api="${control_service_ports_https}" \
```
walterlopatka, line 895: > loadbal.ports.api

Ok, but why not name this `loadbal.ports.https`?

walterlopatka (commented):
Looks good. Minor comment/question, but it won't prevent me from approving.

k8_cortx_cloud/deploy-cortx-cloud.sh
keithpine: Good question.

k8_cortx_cloud/solution.yaml
```diff
@@ -50,7 +50,20 @@ solution:
         sns: 1+0+0
         dix: 1+0+0
     external_services:
-      type: LoadBalancer
+      s3:
```
osowski, line 53: Can you make these same changes in the solution_stub.yaml as well? This will ensure we can keep them in sync and not have stub deployments fail validation.

osowski (changes requested):
Looks good to me, with the exception we'll want to make the same modifications to solution_stub.yaml.

osowski (approved):
lgtm

PR #124, opened 2022-02-28
Fixing broken links and improving them

johnbent:
First, the minikube link was 404. While fixing that I noticed that both links were using absolute paths. Relative are better because they resolve appropriately within a branch. Using an absolute link means you have to choose which branch and embed that in the URL.

Signed-off-by: John Bent <john.bent@seagate.com>

-----
[View rendered README.md](https://github.com/Seagate/cortx-k8s/blob/johnbent-patch-1/README.md)


osowski (approved):
lgtm

PR #125, opened 2022-03-01
CORTX-28899: Ability to create multiple LoadBalancer IO services

keithpine:
The `solution.yaml` and platform Helm chart have been updated to support the ability to create multiple CORTX IO Services. This service provides access to the RGW servers for S3 IO. By default, only one service is created. Multiple services can be created for redundancy in a cloud environment.

- Multiple service instances are only supported for the `LoadBalancer` service type. Defining multiple `ClusterIP` or `NodePort` services doesn't provide any benefit, so those are fixed to a single instance.
- If custom nodePorts are configured, the first instance of the service will take the specified nodePort value, and the port number will be incremented for each additional service instance. The port numbers must not conflict else the deployment will fail.
- The service names are assigned based on the Chart value and a sequential incrementing number, e.g. `cortx-io-svc-0`, `cortx-io-svc-1`, ..., `cortx-io-svc-N`.

**Possible breaking changes**

- Solution configuration field `solution.common.external_services.s3.count` is required. Set to 1 for previous behavior.
- If you previously had a dependency on the service name, then you will need to update to use the generated number-based names instead, e.g. `cortx-io-svc-0` instead of `cortx-io-svc`.

**TODO**
The configmap (config.yaml) makes use of the service name. For now this is fixed to the singular `cortx-io-svc-0`. We need to understand what this value is being used for. For example, the CSM agent endpoint uses this name, which doesn't make much sense.

### DIFF
--- a/k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-configmap/templates/_config.tpl
+++ b/k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-configmap/templates/_config.tpl
@@ -1,4 +1,5 @@
 {{- define "config.yaml" -}}
+{{- $ioSvcName := printf "%s-0" .Values.cortxIoServiceName -}}
 cortx:
   external:
     {{- if .Values.externalKafka.enabled }}
@@ -40,15 +41,15 @@ cortx:
   s3:                                                               # DEPRECATED - ENTIRE S3 KEY
     iam:
       endpoints:
-      - https://{{ .Values.cortxIoServiceName }}:9443
-      - http://{{ .Values.cortxIoServiceName }}:9080
+      - https://{{ $ioSvcName }}:9443
+      - http://{{ $ioSvcName }}:9080
     data:
       endpoints:
-      - http://{{ .Values.cortxIoServiceName }}:80
-      - https://{{ .Values.cortxIoServiceName }}:443
+      - http://{{ $ioSvcName }}:80
+      - https://{{ $ioSvcName }}:443
     internal:
       endpoints:
-      - http://{{ .Values.cortxIoServiceName }}:28049
+      - http://{{ $ioSvcName }}:28049
     {{- with .Values.cortxS3.instanceCount }}
     service_instances: {{ . | int }}
     {{- end }}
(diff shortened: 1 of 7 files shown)


k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-configmap/templates/_config.tpl
```diff
@@ -62,16 +63,16 @@ cortx:
...
       endpoints:
-      - http://{{ .Values.cortxIoServiceName }}:8000
-      - https://{{ .Values.cortxIoServiceName }}:8443
+      - http://{{ $ioSvcName }}:8000
+      - https://{{ $ioSvcName }}:8443
     s3:
       endpoints:
-      - http://{{ .Values.cortxIoServiceName }}:8000
-      - https://{{ .Values.cortxIoServiceName }}:8443
+      - http://{{ $ioSvcName }}:8000
```
keithpine, line 74: Need clarification on whether the service name is used for _any_ of these endpoints. In other words, what do these settings mean?

k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-platform/templates/cortx-io-svc.yaml
```diff
@@ -1,26 +1,32 @@
...
     cortx.io/service-type: "cortx-server"
   ports:
-    {{- $nodePortAllowed := or (eq $svc.type "NodePort") (eq $svc.type "LoadBalancer") }}
     - name: cortx-rgw-http
       protocol: TCP
       port: {{ $svc.ports.http }}
       targetPort: 8000
       {{- if and $nodePortAllowed (not (empty $svc.nodePorts.http)) }}
-      nodePort: {{ $svc.nodePorts.http }}
+      nodePort: {{ add $svc.nodePorts.http $i }}
```
keithpine, line 22: Alternatively, the nodePorts could be a list of nodePorts with length `.Values.services.io.count`. I've seen it specified this way in other charts.

osowski: I think I like it better the way you have it implemented here at the moment. It aligns with the rest of our patterns for `starting_port + count` to determine sequential port numbers.

osowski (approved):
lgtm.

i will cover the documentation changes for the structural updates to solution.yaml in CORTX-29121.

keithpine:
There are several Codacy warnings worth fixing.

Have you considered using a standard test framework, like [pytest](https://docs.pytest.org/en/7.0.x/), instead of rolling your own?

PR #126, opened 2022-03-01
CORTX-28579:  Basic regression tests for cortx-k8s

walterlopatka:
This PR introduces basic regression tests for cortx-k8s.  The goal of these tests is to implement a smoke test for cortx-k8s scripts and template files.  These tests do not introduce comprehensive system tests for CORTX.

This first release has two tests which follow the same pattern.
* TEST-DEPLOY-0001 deploys using cortx images
* TEST-DEPLOY-0002 deploys using stub (centos7) images

Each test follows this sequence:
1. Run "prereq"
2. deploy-cortx-cloud.sh
3. status-cortx-cloud.sh (confirm everything passes)
4. shutdown-cortx-cloud.sh
5. start-cortx-cloud.sh
6. destroy-cortx-cloud.sh

More checks can be added to this sequence, but at least it exercises all scripts (except upgrade).

-----
[View rendered test/cortx-k8s-tests.md](https://github.com/walterlopatka/cortx-k8s/blob/CORTX-28579_basic_regtest/test/regression/cortx-k8s-tests.md)

### DIFF
--- a/test/regression/checker.py
+++ b/test/regression/checker.py
@@ -1,0 +1,85 @@
+###########################################################
+# checker.py
+#
+# Module to facilitate making specific pass/fail test points
+#
+# Usage:
+# * Create a Checker object.  (If a logger object is not provided
+#     one will be created.
+# * Use "test_pass" or "test_fail" methods to indicate a pass/fail
+# * Use "test" method to log a pass/fail based on the boolean/
+#     conditional that is passed in.
+# * Use "test_equal" to test if a test value is equal to the
+#     expected value.
+#
+# In all cases, the user must provide a message string that will
+# be logged with the result.
+#
+# At then end of a test program, call the result method to
+# print a summary of the results.
+# * Return value 1 indicates at least one test failure
+# * Return value 0 indicates pass
+#
+###########################################################
+
+from utils import Logger
+
+class Checker:
+
+    def __init__(self, logger=None):
+        if not logger:
+            logger = Logger()
+        self.logger = logger
+
+        self.count = 0
+        self.fails = []
+
+
+    def result(self):
+        if self.fails:
+            self.logger.logfail()
+            self.logger.logfail('F '*20)
+            self.logger.logfail(' F'*20)
+            self.logger.logfail()
+            self.logger.logfail('{0} of {1} tests failed'.format(len(self.fails), self.count).center(40))
+            self.logger.logfail()
+            for msg in self.fails:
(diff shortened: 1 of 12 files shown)


test/example_config.yaml
```diff
@@ -0,0 +1,39 @@
+###################################################
+#
+# k8s-cluster.yaml
```
osowski, line 3: As a general comment on this file, so that we don't duplicate effort in understanding CORTX as a whole and eventual multiple documentation domains, would we be able to align the structure of this file with what is in solution.yaml in a more 100% compatible way, such that you could in theory do a `yq merge` operation and overlay these values onto the default `solution.yaml`?

osowski (changes requested):
Can we move the bulk of the regression work to a subdir under the new `test` directory you create here? IE Create `test/regression/checker.py` instead of just `test/checker.py`, as I'd imagine more "test" projects could live under the `/test` subdir than just the basic regression. 

Also made one formatting / structural comment on the example_config.yaml file to stimulate conversation.

test/example_config.yaml
walterlopatka: I'm open to doing that, but one of the goals is to make it simple to configure.

For cvgs I find it annoying and verbose to name each of the data devices, to specify the size, and to specify "type: ios" (I don't know what it does).

For nodes I find it annoying to use the keys "node1", "node2", rather than just list the set of hostnames.

walterlopatka: Regarding putting these in a "regression" directory, good idea!  I'm happy to do that.

keithpine:
There are several Codacy warnings worth fixing.

Have you considered using a standard test framework, like [pytest](https://docs.pytest.org/en/7.0.x/), instead of rolling your own?

walterlopatka:
> There are several Codacy warnings worth fixing.
> 
> Have you considered using a standard test framework, like [pytest](https://docs.pytest.org/en/7.0.x/), instead of rolling your own?

I will fix those Codacy issues.
Regarding a framework, I'm ambivalent.  I just want something that will run arbitrary programs and print results in a readable format.  And then I add a few modules to commonly repeated things.  I'm open to improving here if it looks to be valuable in the future.

keithpine:
You might want to rebase this PR, it seems you've pulled in already merged commits.

keithpine (approved):

test/example_config.yaml
walterlopatka: I have moved this content from `test` to `test/regression` as requested.

PR #128, opened 2022-03-01
CORTX-29126: Remove status-cortx-cloud.sh from checking svc type

walterlopatka:
Previously, status-cortx-cloud.sh assumed that the service type
for the cortx-control-loadbal-svc and cortx-server-loadbal-svc
was "LoadBalancer".  This is no longer the case.  In fact, the default
type is now "NodePort".  This is not a useful status check anyway,
so this changset removes the check for the service type.


walterlopatka:
PR #129 covers these changes.  I am closing this PR.

PR #129, opened 2022-03-01
CORTX-29125: update status-cortx-cloud.sh to return exit code 1 on failure

walterlopatka:
Two Issus are addressed by this PR.

**CORTX-29125: Update status-cortx-cloud.sh to return exit code 1 on failure.**
Previously status-cortx-cloud.sh always returned 0.  This PR brings two changes:
1. On failure this script returns 1
2. At the end of the script it prints a summary.  Either:
  a. A note that all status checks passed
  b. A count of status checks that failed

**CORTX-29126: Do not check for LoadBalancer service type**
Previously, status-cortx-cloud.sh assumed that the service type
for the cortx-control-loadbal-svc and cortx-server-loadbal-svc
was "LoadBalancer". This is no longer the case. In fact, the default
type is now "NodePort". This is not a useful status check anyway,
so this changset removes the check for the service type.

### DIFF
--- a/k8_cortx_cloud/status-cortx-cloud.sh
+++ b/k8_cortx_cloud/status-cortx-cloud.sh
@@ -9,6 +9,8 @@ then
     exit 1
 fi
 
+failcount=0
+
 FAILED='\033[0;31m'       #RED
 PASSED='\033[0;32m'       #GREEN
 ALERT='\033[0;33m'        #YELLOW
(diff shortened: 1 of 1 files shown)


walterlopatka:
I would prefer not to update the issues that Codacy flagged.  The printf usage that was flagged is consistent with the dozens of other uses for printf in this script.

osowski (approved):
lgtm

osowski:
@walterlopatka  This looks like it includes https://github.com/Seagate/cortx-k8s/pull/128 already... unless I am mis-reading something. So we don't need the explicit https://github.com/Seagate/cortx-k8s/pull/128 PR, do we?

walterlopatka:
Yes, that's right.   Let me just update the PR message here to reflect both, and I will close #128.  (I generally like to keep everything separate, but I was a little concerned about merges, etc., so I just built one on top of the other.  Probably not great, but ok for these simple changes.)

walterlopatka:
I have updated the PR message to include both minor issues that this PR addresses.  I will close #128.

keithpine (approved):

PR #130, opened 2022-03-02
CORTX-29144: Fix shellcheck issues in logs-cortx-cloud.sh

keithpine:
- This { is literal. Check expression (missing ;/\n?) or quote it. [SC1083]
- Consider invoking this command separately to avoid masking its return value (or use '|| true' to ignore). [SC2312]
- Quote to prevent word splitting/globbing, or split robustly with mapfile or read -a. [SC2206]
- In [[..]], use && instead of -a. [SC2108]
- Prefer [[ ]] over [ ] for tests in Bash/Ksh. [SC2292]
- Expanding an array without an index only gives the first element. [SC2128]

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -29,5 +29,17 @@ jobs:
       - uses: actions/checkout@v2
       - name: Run Shellcheck
         uses: Seagate/action-shellcheck@1.1.0
-        # Temporarily pass until files are fixed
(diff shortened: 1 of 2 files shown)


keithpine:
Codacy result:

```
Issues
======
+ Solved 42
```

walterlopatka (approved):
lgtm

osowski (approved):
lgtm

PR #136, opened 2022-03-03
CORTX-28870: Adding the allowable list for Alex scan 

venkuppu-chn:
Signed-off-by: Venkatesh K <venkatesh.k@seagate.com>

**Problem Statement:**
Alex scan generate false positive

**Solution:**
We have added some allowable list to ignore the false positives.


osowski (approved):
lgtm

PR #141, opened 2022-03-04
CORTX-28036: Remove OpenLDAP from CORTX deployment

walterlopatka:


k8_cortx_cloud/solution.yaml
```diff
@@ -3,11 +3,10 @@ solution:
   secrets:
     name: cortx-secret
     content:
-      openldap_admin_secret: seagate1
       kafka_admin_secret: Seagate@123
       consul_admin_secret: Seagate@123
       common_admin_secret: Seagate@123
-      s3_auth_admin_secret: ldapadmin
+      s3_auth_admin_secret: cortxadmin
```
walterlopatka, line 9: I don't know why the default password is `ldapadmin`.  I am changing it to the more reasonable `cortxadmin`.

k8_cortx_cloud/destroy-cortx-cloud.sh
```diff
@@ -290,28 +289,6 @@ function deleteKafkaZookeper()
     helm uninstall zookeeper -n "default"
 }
 
-function deleteOpenLdap()
```
keithpine, line 293: I think you would want to keep this and the other stuff for a couple releases. Someone who uses this script on an existing cluster with openldap won't have it cleaned up.

There are other examples which do the same, and ignore errors if not missing. I would look to them.

keithpine: Just ran into this actually, making a patch on top of this PR and deleting the cluster stalled when deleting PVCs. The script hung until I manually uninstalled the chart (and the pods deleted).

walterlopatka: I agree.  (And I saw the same thing you did, once I started testing.)  I have updated this PR so that it will silently delete openldap resources if they are present (statefulset, pvcs, pvs, services).  It's not completely silent -- the helm uninstall leaves a little message.  But for the normal case where openldap is not installed, it silently skips over it.

Thanks for the comment... this is a good change.

(This also gives me ideas for additional regression tests.  It should be straightforward to script up deploying / destroying different versions.)

k8_cortx_cloud/destroy-cortx-cloud.sh
```diff
@@ -451,7 +455,7 @@ function helmChartCleanup()
     printf "########################################################\n"
     print_header=true
     local charts
-    charts="$(helm list | { grep '^consul\|^cortx\|^kafka\|^openldap\|^zookeeper' || true; })"
+    charts="$(helm list | { grep '^consul\|^cortx\|^kafka\|^openldap|^zookeeper' || true; })"
```
osowski, line 458: Missing a slash between `openldap` and `zookeeper`.

osowski: The rest of the PR looks good to me. I will merge and patch this in a subsequent PR due to the source here being in Walter's forked repo.

osowski (approved):
lgtm.

I will patch the one missing item in a follow-up PR due to this being in Walter's forked repo.

k8_cortx_cloud/destroy-cortx-cloud.sh
keithpine:
```suggestion
    charts="$(helm list | { grep '^consul\|^cortx\|^kafka\|^openldap\|^zookeeper' || true; })"
```

keithpine (approved):

PR #143, opened 2022-03-07
CORTX-29144: Fix shellcheck issues in status-cortx-cloud.sh

keithpine:
Fix the following shellcheck errors/warnings:

- Consider invoking this command separately to avoid masking its return value (or use '|| true' to ignore). [SC2312]
- Expanding an array without an index only gives the first element. [SC2128]
- Useless echo? Instead of 'echo $(cmd)', just use 'cmd'. [SC2005]
- node_list_str appears unused. Verify use (or export if used externally). [SC2034]
- Prefer [[ ]] over [ ] for tests in Bash/Ksh. [SC2292]
- $/${} is unnecessary on arithmetic variables. [SC2004]

SC2312 is disabled completely in this script since updating each shell statement to return true has little value. The point of the check is to make sure errors are handled, except we don't check for errors at all (or even implicitly by `set -e`).

Other changes:

- Implement color messages according to bash template example (https://betterdev.blog/minimal-safe-bash-script-template/)
- Enable the checking of `status-cortx-cloud.sh` in the GitHub Lint action

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -42,5 +42,4 @@ jobs:
             prereq-deploy-cortx-cloud.sh
             shutdown-cortx-cloud.sh
             start-cortx-cloud.sh
-            status-cortx-cloud.sh
             upgrade-cortx-cloud.sh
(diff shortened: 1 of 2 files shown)


osowski (approved):
lgtm

PR #155, opened 2022-03-10
CORTX-28509: Remove secrets mounts from non-init containers

keithpine:
Secrets are only required in init containers, where the secrets are encrypted and stored during provisioning. Regular containers access the secrets via the cluster config file.


osowski (approved):
lgtm

PR #158, opened 2022-03-11
CORTX-29137: Support setting of some arbitrary config.yaml values

keithpine:
Add `extra_configuration` fields for s3 (rgw) and motr configuration blocks. These are _optional_ multi-line strings that are included in the corresponding `config.yaml` settings.

This allows the configuration of some settings without explicit support in `solution.yaml`:

Example:
```yaml
    s3:
      ...
      extra_configuration: |
        thread_pool_size: 10
        init_timeout: 300
        gc_max_objs: 32
        gc_obj_min_wait: 1800
        gc_processor_max_time: 3600
        gc_processor_period: 3600
        motr_layout_id: 9
        motr_unit_size: 1048576
        motr_max_units_per_request: 8
        motr_max_idx_fetch_count: 30
        motr_max_rpc_msg_size: 524288
```

The above would be appended exactly as specified into the generated ConfigMap value for `config.yaml`, key `s3`. Currently `motr` and `rgw` configuration fields are supported.

Note that this is an advanced configuration application and setting this could render the generated `config.yaml` invalid.


osowski (approved):
lgtm

PR #161, opened 2022-03-12
CORTX-29234: Default nodePorts to `null` in solution.yaml

keithpine:
Cosmetic change only, no functional difference. Since the port value is an integer, `null` makes a little more sense to use than an empty string.

`""`, '~'`, `null` or the omission of the node port values are all valid and have the exact same behavior, which results in the assignment of random port numbers.

-----
[View rendered README.md](https://github.com/keithpine/cortx-k8s/blob/CORTX-29234_nodeport-defaults/README.md)


osowski (approved):
lgtm

PR #164, opened 2022-03-14
Rename `solution.yaml` to `solution.example.yaml`.

keithpine:
This serves a few purposes:

- Less chance of accidental modifications to the source controlled example
- Avoids accidental use of the wrong file (default as `solution.yaml`) when using files with different names
- Makes it obvious that the default file is just an example that should be customized

-----
[View rendered README.md](https://github.com/keithpine/cortx-k8s/blob/solution-rename/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,39 +120,44 @@ If you have direct access to the underlying Kubernetes Nodes in your cluster, CO
 
 1. Clone this repository to a machine with connectivity to your Kubernetes cluster:
 
-   ```bash
-   git clone https://github.com/Seagate/cortx-k8s
-   ```
+    ```bash
+    git clone https://github.com/Seagate/cortx-k8s
+    ```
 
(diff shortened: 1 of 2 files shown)


keithpine:
Could potentially have some impact on CI jobs if they rely on this file.

walterlopatka (approved):
lgtm

keithpine:
@shailesh-vaidya FYI, for Jenkins updates.

osowski (changes requested):
In the spirit of tolerant changes, let's make this a progressive change instead of an abrupt breaking change.

Can we add the new solution.yaml.example file, comment and deprecate the current solution.yaml file, cut a release, make sure processes get updated, then remove the current solution.yaml file in the next release instead.

That should give @shailesh-vaidya and the cortx-re team enough time to adopt changes instead of breaking anything in the interim.

osowski (approved):
lgtm

walterlopatka (approved):

PR #166, opened 2022-03-16
Cortx 29228: removed sleep infinity from csm in control pod & hax in server pod

lakshita-jain:


lakshita-jain:
need to raise pR for cortx-test not integration

PR #172, opened 2022-03-18
CORTX-29179: Update RGW endpoints

keithpine:
Adjust RGW http/https ports to reflect updated CORTX defaults.

- RGW services (container) listen on http/https ports 22751/23001.
- External IO service updated to expose ports 80/443 by default. These are still configurable.
- RGW settings `config.yaml` ConfigMap updated to match latest sample
- S3 settings in `config.yaml` removed as it's no longer used


keithpine:
These changes are not really needed functionality-wise. Aside from hard-coding the ports in the Chart templates, everything is configured correctly. This just brings a new deployment up-to-date with the CORTX defaults, and cleans up the `cluster.yaml` file.

k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-configmap/templates/_config.tpl
```diff
@@ -30,53 +29,29 @@ cortx:
...
-    service_instances: {{ . | int }}
-    {{- end }}
-    io_max_units: 8
-    {{- with .Values.cortxS3.maxStartTimeout }}
-    max_start_timeout: {{ . | int }}
-    {{- end }}
+  rgw:
+    {{- $ioSvcName := required "A valid cortxIoService.name is required!" .Values.cortxIoService.name }}
+    {{- $iosvcHttpPort := required "A valid cortxIoService.ports.http is required!" .Values.cortxIoService.ports.http }}
+    {{- $iosvcHttpsPort := required "A valid cortxIoService.ports.https is required!" .Values.cortxIoService.ports.https }}
```
keithpine, line 35: This could be achieved by a Helm `lookup`, however the deployment script installs this chart prior to the cortx-platform chart, so the service is not available.

k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-configmap/templates/_config.tpl
```diff
@@ -30,53 +29,29 @@ cortx:
...
     auth_user: {{ .Values.cortxRgw.authUser }}
     auth_admin: {{ .Values.cortxRgw.authAdmin }}
     auth_secret: {{ .Values.cortxRgw.authSecret }}
-  rgw:
-    iam:                                                            # DEPRECATED - IAM KEY
+    s3:                                                        # deprecated
```
keithpine, line 39: Despite being marked as `deprecated`, Provisioner will crash on setup if this is removed.

k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-server/templates/cortx-server-pod.yaml
```diff
@@ -120,8 +120,8 @@ spec:
             - name: UDS_CLOUD_CONTAINER_NAME
               value: {{ .Values.cortxserver.name }}
           ports:
-          - containerPort: 8000
-          - containerPort: 8443
+          - containerPort: 22751
```
osowski, line 123: We don't have to do it now, but if we add a name to these ports, then that name can be referenced in the Service's corresponding targetPort field in `cortx-io-svc.yaml` instead of needing to manually keep the ports in sync (so that way the names are only written once and the service automatically tracks the containerPort changes in the Pod spec).

k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-server/templates/cortx-server-svc.yaml
```diff
@@ -37,8 +37,8 @@ spec:
     - name: cortx-rgw-http
       protocol: TCP
       port: {{ .Values.cortxserver.service.loadbal.ports.http }}
```
osowski, line 39: Are we missing the updates to the individual NodePort Services via `.Values.cortxserver.service.loadbal.ports.http` that reflect the same intent as the `cortx-io-svc` (configured via cortx-servers's values.yaml file)
https://github.com/Seagate/cortx-k8s/blob/b5ba60651caecfebcba672835a76b580a725b47c/k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-server/values.yaml#L22-L23

If I am reading everything else correctly, we'd want to update those defaults to 80 and 443 as well. I believe those are no longer needed from an architectural perspective, but we should keep them in sync with `cortx-io-svc` updates until we can remove them altogether.

osowski (commented):
One question on the updates to the individual Server Pod NodePort Services, but other than that it looks pretty good to me.

k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-server/templates/cortx-server-pod.yaml
keithpine: Good idea. I'll go ahead and add this, I prefer it to manually syncing them.

k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-server/templates/cortx-server-pod.yaml
```diff
@@ -120,8 +120,12 @@ spec:
             - name: UDS_CLOUD_CONTAINER_NAME
               value: {{ .Values.cortxserver.name }}
           ports:
-          - containerPort: 8000
-          - containerPort: 8443
+          - name: rgw-http
+            containerPort: 22751
+            protocol: TCP
```
keithpine, line 125: Superfluous, but it seems conventional to be more explicit...

osowski (approved):
lgtm

PR #175, opened 2022-03-22
CORTX-29179: Remove common.s3.num_inst from solution.yaml validation

walterlopatka:
Remove `num_inst` from solution-check.yaml.


osowski (approved):
lgtm!

PR #176, opened 2022-03-22
CORTX-29122: Restore CORTX passwords

walterlopatka:
We have decided that rather than make such a significant
change as generating passwords by default that we would
first issue a written warning that this will happen in the
next release.

This changeset restores the CORTX passwords that were removed
as part of the prior CORTX-29122 change.  This change
will be reverted in the next release in a few weeks after
the QA and DevOps teams have had a chance to accommodate
these changes.

### DIFF
--- a/k8_cortx_cloud/solution.yaml
+++ b/k8_cortx_cloud/solution.yaml
@@ -6,7 +6,7 @@ solution:
       kafka_admin_secret: Seagate@123
       consul_admin_secret: Seagate@123
       common_admin_secret: Seagate@123
-      s3_auth_admin_secret: cortxadmin
+      s3_auth_admin_secret: ldapadmin
       csm_auth_admin_secret: seagate2
       csm_mgmt_admin_secret: Cortxadmin@123
   images:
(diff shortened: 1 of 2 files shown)


k8_cortx_cloud/solution.yaml
```diff
@@ -3,12 +3,12 @@ solution:
...
-      kafka_admin_secret: null
-      consul_admin_secret: null
-      common_admin_secret: null
-      s3_auth_admin_secret: null
-      csm_auth_admin_secret: null
-      csm_mgmt_admin_secret: null
+      kafka_admin_secret: Seagate@123
+      consul_admin_secret: Seagate@123
+      common_admin_secret: Seagate@123
+      s3_auth_admin_secret: cortxadmin
```
keithpine, line 9: In 0.1.0 this is `ldapadmin`. Any concern that someone may also be hard-coding that one, and switching to `cortxadmin` would be a problem?

osowski (approved):
lgtm

PR #179, opened 2022-03-23
CORTX-29672: Remove unused s3 server settings

keithpine:
Port number 28051 is the S3 Authserver HTTPs port. This is no longer in use with the switch to RGW.


osowski (approved):
lgtm

PR #180, opened 2022-03-23
CORTX-29449: Add cortx-setup initContainer to logs to be collected

walterlopatka:
This PR adds the cortx-setup initContainer to the set of logs that are collected.


walterlopatka:
I have tested this with a normal, deployed cluster.  I haven't yet tested with the case where something fails in init.  I will do that test before I merge.

keithpine (approved):
✨ Nice! ✨

PR #185, opened 2022-03-29
CORTX-29634: Update Consul image version to 1.11.4

walterlopatka:
Update Consul image as specified in solution.example.yaml to 1.11.4.

ghcr.io/seagate has been previously updated to contain this image.

I have verified that CORTX deploys with this new image.


keithpine (approved):
Great! 🌟

PR #188, opened 2022-03-29
CORTX-29435: Issue templates

keithpine:
Add GitHub Issue templates to improve and streamline issue submission and triage.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,38 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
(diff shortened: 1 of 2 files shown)


walterlopatka (commented):
General comment on the bug report:  I don't think the "Desktop" / "Smartphone" / "Browser version" items are relevant for cortx-k8s.  Perhaps these section should be replaced by "environment" with items like "cortx-k8s version", "cortx-image version", and/or "solution.yaml" file.  Maybe even request k8s output (like "describe pod") as necessary.

osowski:
> General comment on the bug report: I don't think the "Desktop" / "Smartphone" / "Browser version" items are relevant for cortx-k8s. Perhaps these section should be replaced by "environment" with items like "cortx-k8s version", "cortx-image version", and/or "solution.yaml" file. Maybe even request k8s output (like "describe pod") as necessary.

Agreed on the overall "Desktop" elements. I think it would be good to have the "Environment" section @walterlopatka noted with:
- Kubernetes Version
- cortx-k8s Version
- cortx container image versions
- (Optional) uploaded solution.yaml (which would negate the need for the container image versions inline)

We may want to establish a mini-"must gather" set of commands that is simpler than running `logs-cortx-cloud` outright, but I don't think that should hold us up here. Something like `kubectl get all && kubectl get nodes` would be simple enough to get an understanding of the deployed runtime environment.

keithpine:
Thanks for the comments. I wasn't ready for any review yet. :D Closing for now.

PR #199, opened 2022-04-01
CORTX-29698: Update regtests for v0.2.0 solution.yaml updates

walterlopatka:
## Description
Update regtests for v0.2.0 solution.yaml updates
* use solution.example.yaml instead of solution.yaml
* add support for setting secrets in config file
* add support for setting nodeports in config file
    (not related to v0.2.0)

## Type of change
- [x] Code quality improvements to existing code or test additions/updates

## How was this tested?
* `./testrunner.py -c 594.yaml -t testlists/deploy_all.py`, where `594.yaml` is my local config with new changes
* Ran special manual tests to be sure that the nodeport solution.yaml update works in all cases

## Checklist

- [ ] All commits are signed off and are in agreement with the [CORTX Community DCO and CLA policy](https://github.com/Seagate/cortx/blob/main/doc/dco_cla.md).

-----
[View rendered test/regression/cortx-k8s-tests.md](https://github.com/walterlopatka/cortx-k8s/blob/CORTX-29698_regression_test_update/test/regression/cortx-k8s-tests.md)

### DIFF
--- a/test/regression/cluster.py
+++ b/test/regression/cluster.py
@@ -57,7 +57,7 @@ class Cluster:
 
         solution = safe_load(open(self.solution_file))
         nodes = solution['solution']['nodes']
-        self.nodes = [nodes[n]['name'] for n in nodes]
+        self.nodes = [n['name'] for n in nodes.values()]
 
     def _verify_cluster_yaml(self):
         if not self.cluster_data['nodes']:
--- a/test/regression/cluster.py
+++ b/test/regression/cluster.py
@@ -57,8 +57,7 @@ class Cluster:
 
         solution = safe_load(open(self.solution_file))
         nodes = solution['solution']['nodes']
-        self.nodes = [ nodes[n]['name'] for n in nodes ]
-
+        self.nodes = [nodes[n]['name'] for n in nodes]
 
     def _verify_cluster_yaml(self):
         if not self.cluster_data['nodes']:
@@ -86,8 +85,6 @@ class Cluster:
 
     def generate_solution_file(self):
 
-        cortx_ver = self.cluster_data.get('cortx_ver')
-        setup_size = self.cluster_data.get('setup_size')
         generated_solution = self.cluster_data.get('generated_solution')
 
         if not generated_solution:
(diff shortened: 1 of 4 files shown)


test/regression/cluster.py
```diff
@@ -115,6 +107,34 @@ def generate_solution_file(self):
...
+        #
+        # Set secrets
+        #
+        secrets = self.cluster_data.get('secrets')
+        if secrets:
+            solution['solution']['secrets'] = secrets
+
+        #
+        # Set nodeports
+        #
```
walterlopatka, line 119: This isn't beautiful, but it allows any or all fields in "nodePorts" to be set.

test/regression/cluster.py
```diff
@@ -57,8 +57,7 @@ def __init__(self, solution_file, cluster_file=None, logger=None, logdir=None):
 
         solution = safe_load(open(self.solution_file))
         nodes = solution['solution']['nodes']
-        self.nodes = [ nodes[n]['name'] for n in nodes ]
-
+        self.nodes = [nodes[n]['name'] for n in nodes]
```
keithpine, line 60:
```suggestion
        self.nodes = [n['name'] for n in nodes.values()]
```

test/regression/cluster.py
keithpine: Since all of the nodePorts settings are optional, why do you need to go through with all of this?

walterlopatka: I am defining a minimal config file so that I can easily generate a solution.yaml file.  I don't want to have to specify the parts that don't change -- only the node ports.  (I admit that I'm not sure that I need this, but I will use it for now.  I don't want to use default node ports.).

test/regression/cluster.py
walterlopatka: done

test/regression/cluster.py
keithpine: Ok, I think it threw me off because you're defining some of the ports (as `None`), even when they aren't specified by the minimal config.

If you're not validating the schema, you could just do:

```python
extsvc = solution['solution']['common']['external_services']
extsvc.update(nodeports)
```

I guess I'm not seeing the value of `new_nodeports` either way. Why not just modify the solution config as you go?

```python
extsvc = solution['solution']['common']['external_services']
if nodeports:
    if 'control' in nodeports and 'https' in nodeports['control']:
        extsvc.update({'control': { 'https': nodeports['control']['https']})
    etc
```

keithpine (approved):

walterlopatka:
I think with the code you have there, the update would overwrite the 'port' and 'type' fields of solution.common.external_services.control.

Either way, I'm going to just merge this PR.  (This might change in the future based on PI-7 work anyway.)

PR #206, opened 2022-04-07
Handle an unset `services.io.count` for the cortx-platform Chart install

keithpine:

## Description
Installing the `cortx-platform` Helm Chart fails if the `services.io.count` value is not explicitly set.

```text
Error: INSTALLATION FAILED: template: cortx-platform/templates/cortx-io-svc.yaml:2:7: executing "cortx-platform/templates/cortx-io-svc.yaml" at <gt $svc.count 0>: error calling gt: incompatible types for comparison
```

It's already defaulted to `0` in `values.yaml`, but I have to assume that value coming from the YAML file is conveyed as a string, so the comparison in the template fails as string and int comparisons aren't legal. The command line set works fine (e.g. `--set services.io.count=1`).

This is fixed by casting the value to an int before comparing.

## Type of change

N/A

## How was this tested?
Installed chart manually. This is the command that would be used for a non-default namespace.

```
helm install cortx-ns-foobar cortx-cloud-helm-pkg/cortx-platform --set namespace.create=true --set namespace.name=foobar
```

## Checklist

### DIFF
--- a/k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-platform/templates/cortx-io-svc.yaml
+++ b/k8_cortx_cloud/cortx-cloud-helm-pkg/cortx-platform/templates/cortx-io-svc.yaml
@@ -1,5 +1,5 @@
 {{- $svc := .Values.services.io -}}
-{{- if gt $svc.count 0 -}}
+{{- if gt (int $svc.count) 0 -}}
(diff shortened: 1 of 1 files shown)


keithpine:
There are other problems so I'll drop this in favor of other fixes.

PR #212, opened 2022-04-14
CORTX-30708: Fix shellcheck issues in solution-validation.sh

keithpine:
## Description

Fix shellcheck checks:
- See if you can use ${variable//search/replace} instead. [SC2001]
- Useless echo? Instead of 'echo $(cmd)', just use 'cmd'. [SC2005]
- Declare and assign separately to avoid masking return values. [SC2155]
- <X> appears unused. Verify use (or export if used externally). [SC2034]
- $/${} is unnecessary on arithmetic variables. [SC2004]

Other:
- Remove trailing whitespace
- Check if solution file exists, return an error if not
- Fix SNS validation check that was wrong due to node count miscalculation

## Type of change

- [X] Code quality improvements to existing code or test additions/updates

## Applicable issues

- This change fixes an issue: CORTX-30708

N/A

## How was this tested?
Ran solution check locally, with expected passing file and various error conditions.

During testing, discovered that the CVG count was not working properly due to an incorrect node count calculation.

## Additional information

Each shellcheck error was fixed in a single commit. Each commit can be reviewed separately to see the changes.

As mentioned, the node counting code is pretty fragile. It will fail for the same reason if, in the future, new keys are added to the nodes entries. Fixing that requires a bit more work and is not done in this PR.

I disabled the checks in each location for SC2001. It's not very easy to try and make Bash pattern globs to replace regular expression, and thus not worth the effort.

## Checklist

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -33,7 +33,6 @@ jobs:
           # Temporarily ignore these until they are fixed
           ignore_paths: >-
             parse_scripts
-            solution_validation_scripts
 
   solution_yaml:
     name: Solution YAML validation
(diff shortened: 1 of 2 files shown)


walterlopatka (approved):
lgtm

PR #214, opened 2022-04-15
CORTX-30781 Add regression test for default vs other namespace

walterlopatka:
## Description
This change adds a feature to the regression test `test_deploy.py` to specify a namespace.
When namespace is specified it overrides the namespace specified in the cluster config file.

In addition, a new test that specifies "default" namespace is added (to ensure that deployment
into the default namespace works as expected).

## Type of change

- [x] Code quality improvements to existing code or test additions/updates

## Applicable issues
- This change is related to an issue: CORTX-30781

## How was this tested?
I have run this via ./testrunner.py -c <config> -t testlists/deploy_all.py

## Checklist

### DIFF
--- a/test/regression/test_deploy.py
+++ b/test/regression/test_deploy.py
@@ -14,8 +14,9 @@ from utils import Logger, StopWatch
 
 def verify_pods_in_namespace(checker, namespace):
     cmd = ['kubectl', 'get', 'pods', '-n', namespace, '--no-headers']
+    # nosec
     stdout = subprocess.Popen(cmd, stdout=subprocess.PIPE) \
-                       .communicate()[0].decode('utf-8')  # nosec
+                       .communicate()[0].decode('utf-8')
     expected_pods = {
         'cortx-control': 0,
         'cortx-data': 0,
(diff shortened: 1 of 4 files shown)


test/regression/test_deploy.py
```diff
@@ -1,13 +1,46 @@
...
 
+import yaml
+
 from checker import Checker
 from cluster import Cluster
 from utils import Logger, StopWatch
 
 
+def verify_pods_in_namespace(checker, namespace):
+    stdout = subprocess.Popen(['kubectl', 'get', 'pods', '-A'], # nosec
```
osowski, line 16: This function is `verify_pods_in_namespace`, but then the command you use to get pods is using the `-A` flag which gets all pods across all namespaces. Should this command be updated to use `-n namespace` instead?

osowski (commented):
Only one comment on usage of `kubectl get pods -A`, but other than that looks good to me.

test/regression/test_deploy.py
walterlopatka: Good point.... not sure why I would have done it this way.  🤔

walterlopatka: Ok, I fixed it.

keithpine (approved):

osowski (approved):
lgtm

PR #256, opened 2022-05-20
temporarily disable Consul serviceaccount patching

keithpine:
## Description
We patch the Consul service accounts to disable auto-mounting of the Kubernetes service tokens. However, this requires a restart of the controllers, which in turns causes havoc as the consul Pods crash and restart (they eventually settle down).

The proper solution is to submit a PR upstream to allow disabling auto-mounting via Helm value so we can turn this back on. Or, as a last resort we could pull in a local version of the Chart, or use a fork.

## How was this tested?
Deployed and noticed there were no consul Pod restarts.

## Additional information
We could make this a script option, but I didn't quite see a need to enable it at this time. If that's necessary, let me know and I can add this functionality.

## Checklist

### DIFF
--- a/k8_cortx_cloud/deploy-cortx-cloud.sh
+++ b/k8_cortx_cloud/deploy-cortx-cloud.sh
@@ -578,19 +578,23 @@ function deployCortx()
         --wait \
         || exit $?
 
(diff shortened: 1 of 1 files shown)


walterlopatka (approved):
👍

osowski (approved):
lgtm

PR #233, opened 2022-05-04
CORTX-30140: Kafka managed as cortx Helm Chart dependency

keithpine:
## Description
Add Kafka as a dependency to the new CORTX chart. The CORTX chart is now optionally responsible, via Helm, for installing Kafka. Follow up to #230.

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [X] New feature (non-breaking change that adds new functionality)
- [ ] Breaking change (bug fix or new feature that breaks existing functionality)
- [ ] Third-party dependency update
- [ ] Documentation additions or improvements
- [ ] Code quality improvements to existing code or test additions/updates

## Applicable issues
- This change fixes an issue: CORTX-30140

## How was this tested?

Deployed a new release using this PR. Performed basic I/O.

Tested backwards compatibility by deploying a release prior to this PR. Then ran all of the updated utility scripts against that deployment (although, see Additional information).

## Additional information

`solution.common.resource_allocation.kafka.log_persistence_request_size` has been removed from the solution config file. Persistence for the log directory is not enabled, so the value is unused.

The `status-cortx-cloud.sh` script will report a failure for Kafka Cluster IP services when running against an earlier deployment, since the names of the services have changed. I don't consider this a breaking change, since the status script should reflect what is expected based on how it is supposed to be deployed with the current version.

## Checklist
- [X] The change is tested and works locally.
- [ ] New or changed settings in the solution YAML are documented clearly in the README.md file.
- [X] All commits are signed off and are in agreement with the [CORTX Community DCO and CLA policy](https://github.com/Seagate/cortx/blob/main/doc/dco_cla.md).

If this change addresses a CORTX Jira issue:

- [X] The title of the PR starts with the issue ID (e.g. `CORTX-XXXXX:`)

-----
[View rendered charts/cortx/README.md](https://github.com/keithpine/cortx-k8s/blob/CORTX-30140_subcharts-kafka/charts/cortx/README.md)

PR #260, opened 2022-05-23
CORTX-30684: change retain policy to Delete for local-path storage

keithpine:
## Description
Change the `reclaimPolicy` of the local path StorageClass to `Delete`, from `Retain`. Note that Delete is the [default](https://kubernetes.io/docs/concepts/storage/storage-classes/#reclaim-policy) policy for dynamic provisioner storage class that doesn't specify any policy, and also the [default](https://github.com/rancher/local-path-provisioner/blob/47ea46952b27e8f60863316af24cc886090326c2/deploy/local-path-storage.yaml#L93) for the Local Path Provisioner.

Using a Retain policy means the cluster administrator needs to manually clean-up the backing data for k8s volumes, after a user deletes their PVCs and PVs. In development usage the local node partitions can easily run out of space without active management, causing deployments to fail. In comparison, if the policy is Delete, the provisioner will automatically remove the backing data on the nodes when the PVCs are deleted (and PVs are unbound). This is the behavior that makes sense as a default for users of the deployment scripts.

It is no longer necessary to manually remove PVs since the provisioner does this itself, in fact doing this at the same time causes the provisioner to be unable to removing the node data.

## Breaking change
It's useful for users to be aware of this change to the default behavior. Running the destroy script will now result in all data being removed from the cluster nodes (although, that is really the intent).

The `destroy-cortx-cloud.sh` script no longer manually removes 3rd party PVs. This means running the script against a cluster where the PV reclaim policies are `Retain`, the PVs will not be removed. Users should manually delete the PVs and data on the cluster nodes.

## Type of change

## Applicable issues

- This change address an issue: CORTX-30684
- This change is related to an issue: CORTX-31536

In debugging 31356 I wanted to ensure that no left over Pod data was influencing repeat runs. Switching to the Delete reclaim policy makes sure the left-over data is automatically cleaned up.

## How was this tested?

Deploy and destroy. I see all node data removed after the destroy.

Saved files into S3 bucket, confirmed after stopping and starting the cluster (stop/start scripts) the data is still accessible.

## Additional information

If users prefer to keep the Retain policy, PVs can be modified after they are created. https://kubernetes.io/docs/tasks/administer-cluster/change-pv-reclaim-policy/

Note that future plans include allowing the use of any StorageClass, not just our specific local-path provisioner. In that case, users are free to define the StorageClass any way they prefer, and clean-up PVs and data as required.

## Checklist

### DIFF
--- a/k8_cortx_cloud/destroy-cortx-cloud.sh
+++ b/k8_cortx_cloud/destroy-cortx-cloud.sh
@@ -198,7 +198,7 @@ function deleteCortxPVs()
         if [[ "${force_delete}" == "--force" || "${force_delete}" == "-f" ]]; then
             kubectl patch pv "${persistent_volume}" -p '{"metadata":{"finalizers":null}}'
         fi
-        kubectl delete pv "${persistent_volume}"
+        kubectl delete pv "${persistent_volume}" --ignore-not-found
     done < <(kubectl get pv --no-headers | grep -E "${pv_filter}" | cut -f1 -d" ")
 }
 
@@ -313,7 +313,7 @@ function delete3rdPartyPVCs()
             kubectl patch pvc --namespace "${namespace}" "${volume_claim}" \
                       -p '{"metadata":{"finalizers":null}}'
         fi
-        kubectl delete pvc --namespace "${namespace}" "${volume_claim}"
+        kubectl delete pvc --namespace "${namespace}" "${volume_claim}" --ignore-not-found
     done < <(kubectl get pvc --no-headers --namespace="${namespace}" | grep -E "${pvc_filter}" | cut -f1 -d " ")
 }
 
@@ -327,7 +327,7 @@ function delete3rdPartyPVs()
         if [[ "${force_delete}" == "--force" || "${force_delete}" == "-f" ]]; then
             kubectl patch pv "${persistent_volume}" -p '{"metadata":{"finalizers":null}}'
         fi
-        kubectl delete pv "${persistent_volume}"
+        kubectl delete pv "${persistent_volume}" --ignore-not-found
     done < <(kubectl get pv --no-headers | grep -E "${pvc_filter}" | cut -f1 -d " ")
 }
 
--- a/k8_cortx_cloud/destroy-cortx-cloud.sh
+++ b/k8_cortx_cloud/destroy-cortx-cloud.sh
@@ -187,21 +187,6 @@ function deleteCortxLocalBlockStorage()
     uninstallHelmChart "cortx-data-blk-data-${namespace}" "${namespace}"
 }
 
-function deleteCortxPVs()
-{
-    printf "########################################################\n"
-    printf "# Delete CORTX Persistent Volumes                      #\n"
-    printf "########################################################\n"
-    local -r pv_filter="^${namespace}/cortx-data-fs-local-pvc|^${namespace}/cortx-control-fs-local-pvc"
-    while IFS= read -r persistent_volume; do
-        printf "Removing %s\n" "${persistent_volume}"
-        if [[ "${force_delete}" == "--force" || "${force_delete}" == "-f" ]]; then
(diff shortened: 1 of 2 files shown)


k8_cortx_cloud/destroy-cortx-cloud.sh
```diff
@@ -198,7 +198,7 @@ function deleteCortxPVs()
         if [[ "${force_delete}" == "--force" || "${force_delete}" == "-f" ]]; then
             kubectl patch pv "${persistent_volume}" -p '{"metadata":{"finalizers":null}}'
         fi
-        kubectl delete pv "${persistent_volume}"
+        kubectl delete pv "${persistent_volume}" --ignore-not-found
```
keithpine, line 201: I don't think it's really necessary to manually delete PVs anymore, but I've left it in. I added these ignore flags since there are race conditions (provisioner deletes PV first) that produce "not found" errors from kubectl. Since the script doesn't check for errors, it sort of pointless, but seemed good to do.

keithpine:
I did a loop of deploy/destroy. I've noticed that some of the data directories are still left over. In the local-path-provisioner log, I see some messages like:

```text
E0523 20:42:58.647690       1 controller.go:991] volume "pvc-94f74a9f-f308-473a-bd88-705ca7f64e5a" in work queue no longer exists
```
`pvc-94f74a9f-f308-473a-bd88-705ca7f64e5a` was the PV which corresponds to a left-over data directory.

In the destroy script, we manually delete the PVs:
```text
Removing pvc-94f74a9f-f308-473a-bd88-705ca7f64e5a
persistentvolume "pvc-94f74a9f-f308-473a-bd88-705ca7f64e5a" deleted
```
My guess is that deleting the volume manually is confusing the provisioner, and it doesn't handle a volume being removed while it's in its work queue.

osowski (dismissed):
based on everything else we've seen and have in plan, this lgtm.

walterlopatka (approved):

k8_cortx_cloud/destroy-cortx-cloud.sh
keithpine: This has been removed.

k8_cortx_cloud/destroy-cortx-cloud.sh
```diff
@@ -187,21 +187,6 @@ function deleteCortxLocalBlockStorage()
     uninstallHelmChart "cortx-data-blk-data-${namespace}" "${namespace}"
 }
 
-function deleteCortxPVs()
```
keithpine, line 190: The calling of this function was removed in #233. I'm not entirely sure why I removed it, although it doesn't seem it was necessary.

osowski:
> I did a loop of deploy/destroy. I've noticed that some of the data directories are still left over. In the local-path-provisioner log, I see some messages like:
> 
> ```
> E0523 20:42:58.647690       1 controller.go:991] volume "pvc-94f74a9f-f308-473a-bd88-705ca7f64e5a" in work queue no longer exists
> ```
> 
> `pvc-94f74a9f-f308-473a-bd88-705ca7f64e5a` was the PV which corresponds to a left-over data directory.
> 
> In the destroy script, we manually delete the PVs:
> 
> ```
> Removing pvc-94f74a9f-f308-473a-bd88-705ca7f64e5a
> persistentvolume "pvc-94f74a9f-f308-473a-bd88-705ca7f64e5a" deleted
> ```
> 
> My guess is that deleting the volume manually is confusing the provisioner, and it doesn't handle a volume being removed while it's in its work queue.

As we're moving towards handling as much k8s functionality as possible, this makes sense to me to remove the manual removal of PVs when we have the underlying StorageClass and dynamic provisioner taking care of some of that. It will help with separation of roles, as long as we're able to still run atomic deploy/destroy loops.

walterlopatka (approved):

osowski (approved):

PR #261, opened 2022-05-24
CORTX-30145: Merge cortx-ha Chart into cortx Chart

keithpine:
## Description

Merge the cortx-ha Helm chart into the unified CORTX Helm Chart. Resources previously deployed by the cortx-ha Chart are now deployed by cortx.

For now, I've simply transplanted the values from the old chart, into the new chart's values under the same `cortxha` key. Some `configmap` values were replaced by these. The templates were updated accordingly. In a future change, I'll be coalescing and re-vamping the rest of the YAML structure as other components are merged. Then it will make more sense as to what the final form of values.yaml should look like.

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [X] New feature (non-breaking change that adds new functionality)
- [ ] Breaking change (bug fix or new feature that breaks existing functionality)
- [ ] Third-party dependency update
- [ ] Documentation additions or improvements
- [ ] Code quality improvements to existing code or test additions/updates

## How was this tested?

- Deployed and un-deployed cluster, include data-only deployment. Results were as expected.
- Ran utility scripts (e.g. status).
- Ran S3 I/O.
- Used `kubectl get -o yaml all,svc,serviceaccount,clusterrole,clusterrolebinding,secret,cm` to capture state of system on current `integration` branch, and then again with this change. Diffed the result. The diff was identical, minus the generated bits, and the expected label and name changes.

## Additional information

As part of the merge, the following additional changes were made:

- Added standard Helm labels to all resources
- Updated existing labels
  - Removed `name: cortx-ha`; `app.kubernetes.io/component: ha` can be used instead if needed
  - Renamed `app: cortx-ha` --> `app.kubernetes.io/component: ha`
- Removed custom volume mount names in Deployment (Pod), change to hard-coded values:
  - `config001` --> `cortx-configuration`
  - `ssl-config001` --> `cortx-ssl-cert`
- Multiple resources now use incorporate the Helm release in their names:
  - ServiceAccount: `ha-monitor` --> `<release>-ha-monitor`
  - ClusterRole: `node-list` --> `<release>-ha-node-list`
  - ClusterRoleBinding: `node-list` --> `<release>-ha-node-list`
  - PVC name: `cortx-ha-fs-local-pvc-cortx` --> `<release>-ha`
  - Service name: `cortx-ha-headless-svc` --> `<release>-ha-headless`
- Replaced secret-info.txt with secretName

[...]

PR #252, opened 2022-05-18
CORTX-30142: Merge cortx-control Chart into cortx Chart

keithpine:
## Description
Merge the cortx-control Helm chart into the unified CORTX Helm Chart. Resources previously deployed by cortx-control are now deployed by cortx.

For now, I've simply transplanted the values from the old chart, into the new chart's values under the same `cortxcontrol` key. The templates were updated accordingly. In a future change, I'll be coalescing and re-vamping the YAML structure as other components are merged. Then it will make more sense as to what the final form of values.yaml should look like.

## Type of change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [X] New feature (non-breaking change that adds new functionality)
- [ ] Breaking change (bug fix or new feature that breaks existing functionality)
- [ ] Third-party dependency update
- [ ] Documentation additions or improvements
- [ ] Code quality improvements to existing code or test additions/updates

## Applicable issues
- This change fixes an issue: CORTX-30142

## How was this tested?

- Deployed and un-deployed cluster, include data-only deployment. Results were as expected.
- Ran utility scripts (e.g. status).
- Ran S3 I/O.
- Used the CSM API to login, create a new user, and get the user info. This demonstrates CSM (Control) services and Pod are working.

## Additional information
While moving the values over from one chart to the other, I did some minor cleanup.

- Removed control PVC name value, this is now based on the release name
- Removed control configmap/ssl-configmap/machine-id name and volume mount name values, these can be static
- The control enabled flag `configmap.cortxControl.enabled`, used for data-only deployments, moved to `cortxcontrol.enabled`.

Other changes:

- Added template helpers to provide configmap and control names
- Deployment `app` labels replaced with `app.kubernetes.io/component`
- Added general `cortx.labels` to Deployment
- Added component label to PVC
- The Deployment, Pod, and Service names/hostnames along with the PVC name are based on the Helm release name
- Replaced secret-info.txt with secretName
- Give a name for the control https container port so the service can reference it by name
- `status-cortx-cloud.sh` now verifies control resources via selectors

## Checklist

- [X] The change is tested and works locally.
- [ ] New or changed settings in the solution YAML are documented clearly in the README.md file.
- [X] All commits are signed off and are in agreement with the [CORTX Community DCO and CLA policy](https://github.com/Seagate/cortx/blob/main/doc/dco_cla.md).

If this change addresses a CORTX Jira issue:

- [X] The title of the PR starts with the issue ID (e.g. `CORTX-XXXXX:`)

-----
[...]

PR #266, opened 2022-05-31
Remove wait when installing CORTX Helm Chart

keithpine:
## Description

This change removes the wait when installing the unified CORTX Helm Chart. Previous to this, the installation would fail if the third party dependencies + HA + Control components do not fully deploy within 5 minutes.

The motivation behind this is that the HA and Control wait timeouts were already effectively removed when merging the Control and HA charts into the unified chart (PRs #252 and #261). We don't know what the right timeout value is for all cases; 5 minutes could be too low or too high depending on the configuration. Instead of adding in yet another variable to enforce a timeout, we prefer that users check the state of their deployment after the script has finished. This can be done with the `status-cortx-cloud.sh` script, or individual `kubectl wait` or `kubectl rollout status` commands. 

Timeouts still exist for checking that:

- the third party services are ready
- all CORTX Data deployments are ready
- all CORTX Server deployments are ready

## Breaking change
No. But the aforementioned timeout variables have already been removed, so we should document them as unused now. Those are: `CORTX_DEPLOY_CONTROL_TIMEOUT` and `CORTX_DEPLOY_HA_TIMEOUT`.

## Type of change

## How was this tested?

Deployed a cluster and performed S3 I/O.

## Additional information

Note that there are existing issues logged for initContainers never completing:

- HA deployment: CORTX-31665
- Control deployment: CORTX-31536

As a result, the deployment script will complete successfully, but these services won't be available. This can be verified with the status script. The workaround is to delete the Pods and let them be re-created.

Removing the Helm install timeout actually improves this situation because it allows the full stack to be deployed without failing part way. The failed Pods can be fixed afterwards.

Also note that it seems possible to perform S3 I/O with the default user w/o the control Pod ever coming up. I'm not sure if there's some setup that was partially done beforehand, or if that's expected...

## Checklist

-----
[View rendered README.md](https://github.com/keithpine/cortx-k8s/blob/no-wait/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,14 +343,12 @@ The Helm charts work with both "stub" and "CORTX ALL" containers, allowing users
 
 ### Overriding Helm Install  / Wait Timeouts
 
-There is a "wait" after each of the cortx helm charts are deployed.  This wait is guarded by a timeout.  If needed, these timeout values can be overridden by environment variables.
+There is a "wait" after some of the CORTX Helm Charts are deployed.  This wait is guarded by a timeout.  If needed, these timeout values can be overridden by environment variables.
 
 | Environment Variable           | Default Value |
 | ------------------------------ | ------------- |
-| `CORTX_DEPLOY_CONTROL_TIMEOUT` | `300s`        |
 | `CORTX_DEPLOY_DATA_TIMEOUT`    | `300s`        |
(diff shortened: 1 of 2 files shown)


keithpine:
Closing in favor of #267.

PR #273, opened 2022-06-04
Bump VERSION to v0.8.0.dev0

keithpine:
## Description

Bump version to v0.8.0.dev0.

We haven't been keeping this file up to date in the integration branch (it's not used for anything), although we should so it clearly indicates that it's a development build of the next version. I'm not sure if this is the exact format we want. It depends on how closely we want to follow semver. Other choices could be `v0.8.0-dev` or `v0.8.0-alpha`. I haven't found much info about how semver handles dev versions, if it does at all. This version is based on Python's equivalent.

Semver: `1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0`
PEP 0440: `.devN, aN, bN, rcN, <no suffix>, .postN`

https://semver.org/
https://peps.python.org/pep-0440/


walterlopatka (approved):
Good move!  (I'll add to the release process the creation of this new tag on integration branch for the future.)

PR #276, opened 2022-06-07
CORTX-31831: Add Kafka init container to wait for Zookeeper

walterlopatka:
## Description
Use an init container to prevent Kafka containers to start before Zookeeper is ready for them.  This prevents Kafka
from doing a restart right out of the box.  This is does not fix any functional problem (it is ok if the Kafka container
restarts), but it looks nicer and better controlled.

## Type of change

## Applicable issues

- This change fixes an issue: CORTX-31831

N/A

## How was this tested?
Deployed cluster locally multiple times.  Confirmed no Kafka restarts.

## Checklist

### DIFF
--- a/charts/cortx/values.yaml
+++ b/charts/cortx/values.yaml
@@ -61,6 +61,11 @@ kafka:
   # -- Overridden min.insync.replicas config for the transaction topic
   transactionStateLogMinIsr: 2
 
+  initContainers:
+  - name: zookeeper-watcher
+    image: ghcr.io/seagate/busybox:latest
+    command: ['sh', '-c', 'until echo "ruok" | timeout 30 nc -w 30 cortx-zookeeper 2181 | grep imok; do sleep 1; done']
+
   # ZooKeeper chart configuration
   # ref: https://github.com/bitnami/charts/blob/master/bitnami/zookeeper/values.yaml
   zookeeper:
--- a/charts/cortx/values.yaml
+++ b/charts/cortx/values.yaml
@@ -62,7 +62,7 @@ kafka:
   transactionStateLogMinIsr: 2
 
   initContainers:
-  - name: zookeeper-watcher
+  - name: wait-for-zookeeper
     image: ghcr.io/seagate/busybox:latest
     command: ['sh', '-c', 'until echo "ruok" | timeout 30 nc -w 30 cortx-zookeeper 2181 | grep imok; do sleep 1; done']
 


charts/cortx/values.yaml
```diff
@@ -61,6 +61,11 @@ kafka:
   # -- Overridden min.insync.replicas config for the transaction topic
   transactionStateLogMinIsr: 2
 
+  initContainers:
+  - name: zookeeper-watcher
```
keithpine, line 65:
```suggestion
  - name: wait-for-zookeeper
```

Nitpick, but I prefer the "wait-for" naming that Bitnami uses, it does what is says. "watcher" makes me think of some kind of continuous monitor.

charts/cortx/values.yaml
```diff
@@ -61,6 +61,11 @@ kafka:
...
+    image: ghcr.io/seagate/busybox:latest
+    command: ['sh', '-c', 'until echo "ruok" | timeout 30 nc -w 30 cortx-zookeeper 2181 | grep imok; do sleep 1; done']
```
keithpine, line 67: You can't hard-code the service name. The [name of the service](https://github.com/bitnami/charts/blob/dfbf28240eca7b8732d83d07f10d7a3ca0788e3a/bitnami/zookeeper/templates/svc.yaml#L4) is derived from the Helm release name (or overridden values), and technically the port number is [configurable](https://github.com/bitnami/charts/blob/dfbf28240eca7b8732d83d07f10d7a3ca0788e3a/bitnami/zookeeper/values.yaml#L461). See [here](https://github.com/Seagate/cortx-k8s/blob/76b9c9dea71433f79b3b518ee5f1405431b2c047/charts/cortx/templates/control/deployment.yaml#L119) for an example with Kafka on how to templatize these. Zookeeper has the same options. There's also a headless service (e.g. `cortx-zookeeper-headless`), I'm not sure which one makes more sense to use?

Second thing, this initContainer shouldn't be created if an external Kafka cluster is being used. See [here](https://github.com/Seagate/cortx-k8s/blob/76b9c9dea71433f79b3b518ee5f1405431b2c047/charts/cortx/templates/control/deployment.yaml#L83) for an example to omit this if the Chart is not enabled.

You can actually test this without a full deployment, just install the cortx chart by hand.

keithpine (changes requested):

charts/cortx/values.yaml
walterlopatka: Ah, yes, of course! When I started intended to figure out how to do this after I demonstrated that the initContainers attribute was working for me at all.  I'll definitely update this as you suggest.
Thanks for the suggestions.

walterlopatka:
This was addressed by a different method in #312

PR #277, opened 2022-06-07
Fix status script for StatefulSet, label updates, etc.

keithpine:
- Update status script to handle StatefulSet instead of Deployment, and to recognize number of instances deployed.
- Replace `app` label with `app.kubernetes.io/component`.
- Add component label to service metadata so it can be looked up via selector
- Rename pvc from `local-path` to `data`. This aligns with the convention of other public charts. Plus, a user doesn't necessarily need to use "local path" storage.
- Drop unused Chart values

A handful of these changes are in preparation for the eventual movement to the unified Chart.


keithpine:
Example of PVC name change.

```text
❯ k get pvc -l app.kubernetes.io/component=server
NAME                  STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
data-cortx-server-0   Bound    pvc-7e7c4280-c737-4e49-b1ec-757c8e9e8e2b   1Gi        RWO            local-path     9m7s
data-cortx-server-1   Bound    pvc-c7bcec15-3067-4f38-af9e-e0f631fcb67a   1Gi        RWO            local-path     9m7s
data-cortx-server-2   Bound    pvc-e57e4c73-1029-4965-9804-4e523b82b91f   1Gi        RWO            local-path     9m7s
data-cortx-server-3   Bound    pvc-0137af23-86dd-4f0f-ba08-7b4216f55be0   1Gi        RWO            local-path     9m7s
```
`server` is admittedly pretty generic, but it goes along with "-ha" and "-control". Is there a better name?

osowski (approved):
lgtm

PR #278, opened 2022-06-07
Use selectors for StatefulSet lookups

keithpine:
Use new selector to get statefulset and pod lists.


keithpine:
Not really a necessary change since the problem was fixed elsewhere, but it's a good a time as any to switch to using selectors instead of grep, since the resource type has also changed.

osowski (approved):
lgtm

PR #290, opened 2022-06-20
CORTX-30147: misc Chart-related cleanup

keithpine:
## Description

Various Helm Chart improvements.

- Remove generation of Server endpoints from deploy script. This is done in the Helm templates now.
- Update all port definitions:
  - Port numbers to match the current [sample file](https://github.com/Seagate/cortx-prvsnr/blob/1c0926ec0171eddf761bdaae13c91a4f336c288b/conf/solution/config.yaml.sample) and [Ports page](https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/791216223/CORTX+Security+Ports+needed+by+CORTX+services).
  - Use templates to define the numbers in a single location
  - Add names to all containerPorts, and use in related services
- Simplify local block storage class naming

## Type of change

- [X] Code quality improvements to existing code or test additions/updates

## Applicable issues

- This change fixes an issue: CORTX-30147

## How was this tested?

Deploy/destroy cluster, run util scripts, S3 I/O, CSM API requests.

## Checklist

-----
[View rendered charts/cortx/README.md](https://github.com/keithpine/cortx-k8s/blob/CORTX-30147_endpoint-cleanup/charts/cortx/README.md)

### DIFF
--- a/charts/cortx/README.md
+++ b/charts/cortx/README.md
@@ -105,7 +105,7 @@ helm uninstall cortx
 | cortxcontrol.machineid.value | string | `""` |  |
 | cortxcontrol.service.loadbal.enabled | bool | `true` |  |
 | cortxcontrol.service.loadbal.nodePorts.https | string | `""` |  |
-| cortxcontrol.service.loadbal.ports.https | int | `23256` |  |
+| cortxcontrol.service.loadbal.ports.https | int | `8081` |  |
 | cortxcontrol.service.loadbal.type | string | `"NodePort"` |  |
 | cortxcontrol.sslcfgmap.mountpath | string | `"/etc/cortx/solution/ssl"` |  |
 | cortxdata.blockDevicePaths | list | `[]` |  |
@@ -124,7 +124,6 @@ helm uninstall cortx
 | cortxdata.motr.resources.limits.memory | string | `"2Gi"` |  |
 | cortxdata.motr.resources.requests.cpu | string | `"250m"` |  |
 | cortxdata.motr.resources.requests.memory | string | `"1Gi"` |  |
-| cortxdata.motr.startportnum | int | `29000` |  |
 | cortxdata.nodes | list | `[]` |  |
 | cortxdata.persistentStorage.accessModes[0] | string | `"ReadWriteMany"` |  |
 | cortxdata.persistentStorage.volumeMode | string | `"Block"` |  |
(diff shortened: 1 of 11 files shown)


charts/cortx/templates/_config.yaml.tpl
```diff
@@ -13,9 +13,16 @@
...
-{{- $dataHaxPort := 22001 -}}
-{{- $dataIosPort := 21001 -}}
-{{- $dataConfdPort := 22002 -}}
+{{- $serverHostnames := list -}}
+{{- if .Values.cortxserver.enabled -}}
+{{- range $i := until (int .Values.cortxserver.replicas) -}}
+{{- $serverHostnames = append $serverHostnames (printf "%s-%d.%s" (include "cortx.server.fullname" $) $i (include "cortx.server.serviceDomain" $)) -}}
+{{- end -}}
+{{- end -}}
+{{- $haxPort := 22001 -}}
```
osowski, line 22: Can these port numbers be pulled from anywhere (values.yaml) so that they can be uniformly applied and changed only in one place?

keithpine: I was opting not to, since I don't see the point of making it configurable. We could certainly allow it though, if that's preferred.

Or I could make them template functions, which is more verbose, but puts them in one place.

charts/cortx/README.md
```diff
@@ -107,7 +105,7 @@ helm uninstall cortx
 | cortxcontrol.machineid.value | string | `""` |  |
 | cortxcontrol.service.loadbal.enabled | bool | `true` |  |
 | cortxcontrol.service.loadbal.nodePorts.https | string | `""` |  |
-| cortxcontrol.service.loadbal.ports.https | int | `8081` |  |
+| cortxcontrol.service.loadbal.ports.https | int | `23256` |  |
```
osowski, line 108: I don't think this change should be made here, as this would change the port on `cortx-control-loadbal-svc` Service to listen on 23256 and I believe it is still expected to be on 8081 by default. It can forward to 23256 in the container, which you are also doing, but this would change the port the service itself is listening on - which I don't think is desired.

keithpine: To conclude, nothing in the cluster is currently using port 8081, or at least not explicitly. We don't actually write this section to the config.yaml yet, and its purpose isn't clear to me:

```yaml
csm:
    public:                                                    # Public Control Endpoint
      endpoints:
      - https://control-svc:8081
```

But, based on that [sample](https://github.com/Seagate/cortx-prvsnr/blob/325a3e0bb9ef36de6d918d9167ecdabaf4f681d2/conf/solution/config.yaml.sample#L154-L156), we should continue to configure the service port in that way.

charts/cortx/templates/_config.yaml.tpl
keithpine: I'll go with the templates for now, we can always expose them in values.yaml as needed.

osowski (approved):
lgtm

PR #307, opened 2022-07-06
CORTX-32157: Remove support for custom container paths

keithpine:
## Description

This change removes the ability to set custom container paths via the solution.yaml `solution.common.container_path.*` settings.

These paths set the CORTX configuration in `config.yaml` as well as set mount paths in the containers. The problem is that changing these can break the deployment. It either never worked, or broke as the deployment code has evolved. The affected configuration paths `local` and `log`.

The settings are removed from `solution.example.yaml` and a friendly warning is provided to the user notifying them of this change and it suggests removing the settings.

All containers now have hard-coded paths to the previous default and known working path locations. This simplifies the deployment and removes the chance of incorrect user-configuration. Note that the `config` location was never configurable by the user. There are already several hard-coded paths pointing to this location.

When paths are configured to the non-defaults, all CORTX Pods CrashLoop:

- RGW Pods CrashLoop because during init, Hare writes to the `local` path, but RGW setup reads from the `config` path (fetch-fids fails)
- Control Pods CrashLoop during csm agent startup because of the same failure to fetch fid information
- Data and Client Pods CrashLoop because a machine ID isn't found

Changing the `log` path means that log files could be located on ephemeral storage. There is only a single persistent PVC for each Pod, so the logs directory should be located there to preserve logs when Pods are deleted.

## Breaking change
No, but changing the `container_path` settings in solution.yaml no longer has any effect.

## Type of change

## Applicable issues

## How was this tested?

- Ran deployment, status and log scripts.
- Performed S3 I/O and CSM requests.
- Inspected ConfigMap and Workload specifications.

Also deployed prior to this change with different container paths and noticed all the CrashLoops.

## Checklist

- [X] New or changed settings in the solution YAML are documented clearly in the README.md file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,8 +313,6 @@ This section contains common parameters that affect all CORTX components running
 | Name                                                  | Description                                                     | Default Value       |
 | ----------------------------------------------------- | --------------------------------------------------------------- | ------------------- |
 | `common.storage_provisioner_path`                     | TODO       | `/mnt/fs-local-volume` |
-| `common.container_path.local`                         | TODO       | `/etc/cortx` |
-| `common.container_path.log`                           | TODO       | `/etc/cortx/log` |
 | `common.s3.default_iam_users.auth_admin`              | Username for the default administrative user created for internal RGW interactions. Corresponds to `secrets.content.s3_auth_admin_secret` above. | `sgiamadmin` |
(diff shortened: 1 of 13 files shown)


walterlopatka (approved):
lgtm

osowski (approved):
lgtm

PR #318, opened 2022-07-08
CORTX-32157: Refactor some Chart values

keithpine:
## Description

Refactor the values to be more consistent with Helm conventions, reduce duplication, etc. (general cleanup).

- Remove the `configmap` value and relocate/consolidate child values to the top-level, or within their appropriate component sections.
- Default `clusterName` to the release fullname instead of a hard-coded name.
- Hard-code the RGW "auth_secret". This is not configurable as the Secret is created with the exact value. This is consistent with all of the other secrets, and we have never exposed this setting in solution.yaml.
- Set the secret name to an empty default, and flag it as required so the chart install fails if not specified. Existing secret names can also be a Helm template.
- Hard-code the secret's volume name in the Pod specs, there's not much reason to customize it.

## Type of change

## Applicable issues

- This change is related to an issue: #

## How was this tested?

Deployment, ran util scripts, S3 I/O, CSM request, etc.

Since `existingSecret` is required, and is empty by default, a custom "ci" values file was added for chart linting (`ct`).

## Checklist

### DIFF
--- a/charts/cortx/README.md
+++ b/charts/cortx/README.md
@@ -61,25 +61,8 @@ helm uninstall cortx
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | clusterDomain | string | `"cluster.local"` | Kubernetes Cluster Domain |
-| configmap.clusterDomain | string | `"cluster.local"` |  |
-| configmap.clusterId | string | `""` |  |
-| configmap.clusterName | string | `"cortx-cluster"` |  |
-| configmap.clusterStorageSets | object | `{}` |  |
-| configmap.cortxControl.agent.resources.limits.cpu | string | `"500m"` |  |
-| configmap.cortxControl.agent.resources.limits.memory | string | `"256Mi"` |  |
-| configmap.cortxControl.agent.resources.requests.cpu | string | `"250m"` |  |
-| configmap.cortxControl.agent.resources.requests.memory | string | `"128Mi"` |  |
-| configmap.cortxMotr.confd.resources.limits.cpu | string | `"500m"` |  |
-| configmap.cortxMotr.confd.resources.limits.memory | string | `"512Mi"` |  |
-| configmap.cortxMotr.confd.resources.requests.cpu | string | `"250m"` |  |
-| configmap.cortxMotr.confd.resources.requests.memory | string | `"128Mi"` |  |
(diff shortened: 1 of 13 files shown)


keithpine:
@tanujashinde0405 I'm not aware of any documentation in regards to the allowed values for the `cluster.yaml` file's `cluster.name` and `cluster.id` fields. Could you confirm whether there are any character or length restrictions? As far as I can tell, there isn't. The Chart Release fullname is limited to 63 characters, with no trailing `-`.

https://github.com/Seagate/cortx-k8s/blob/153fa5700cace4e7ab0f74c4bcc11c415f086143/charts/cortx/templates/_helpers.tpl#L8-L24

tanujashinde0405:
> @tanujashinde0405 I'm not aware of any documentation in regards to the allowed values for the `cluster.yaml` file's `cluster.name` and `cluster.id` fields. Could you confirm whether there are any character or length restrictions? As far as I can tell, there isn't. The Chart Release fullname is limited to 63 characters, with no trailing `-`.
> 
> https://github.com/Seagate/cortx-k8s/blob/153fa5700cace4e7ab0f74c4bcc11c415f086143/charts/cortx/templates/_helpers.tpl#L8-L24

@nitin-seagate Can you please confirm this?

nitin-seagate:
> onfirm whether there are any character or length restrictions? As far as I can tell, there isn't. The Chart Release fullname is limited to 63 characters, with no trailing

Yes there is no specification for allowed values wrt Chart Release fullname, we can assume default character-set (alphanumeric with limited special characters)... The cluster ID should be generic uuid (we can govern and defined format for the same in-line with what we followed in LR)

tanujashinde0405:
> 

IN LR also Cluster-id was a uuid we used to generate during deployment : https://github.com/Seagate/cortx-prvsnr/blob/legacy-main/api/python/provisioner/commands/cluster_id.py#L87

keithpine:
Reverted the change to the default cluster ID values. However, previously the dashes `-` were removed from the generated UUD, but now those are kept, so now it is a real UUIDv4, in the same way the LR Provisioner generated it.

I'm not sure it makes sense from a Core perspective that the ID should be required to be a UUID, especially since the software is allowing regular strings. But that's a discussion for another time and place.

PR #327, opened 2022-07-14
CORTX-32157: Refactor Data chart values

keithpine:
## Description

Cleanup Data chart values, in the same vein as previous changes.

* Rename object from cortxdata to data
* Use camel case for variable names
* Use new image style
* Etc.

This has significant changes in regards to the storage set definitions. The storage sets are now defined as a top level list, in the exact same format as [solution.yaml](https://github.com/Seagate/cortx-k8s/blob/6f7cf87636f0020e63f9fda4703914a83f437375/k8_cortx_cloud/solution.example.yaml#L163-L216) except for:

- Key names use camel case
- The `nodes` key is removed since it is not being used. Instead, the `data.replicaCount` defines the number of nodes

The logic to construct the CVG groups and determine the number of StatefulSets has been changed to use Helm (Sprig's really) `chunk` function, which groups all the CVG objects into the storage set's `containerGroupSize` value. This grouping means less manual calculation of counts, and no need for tracking index values.

There is still only support for a single storage set, the template errors out if there's more than one.

The block device PV access modes have been changed to `ReadWriteOnce` by default. `ReadWriteMany` isn't necessary.

Port definitions have been removed from the data headless Service. It was not worth the effort to properly handle the dynamic nature of the CVGs, and none of the other headless services specify anything. The port definitions are merely documentation not functional. We can decide to add them back later if necessary.

## Type of change

## Applicable issues

## How was this tested?

Deployment on a 2-node cluster with two CVGs, and using both group sizes of 1 and 2, with S3/IO. Ran status script.

I capture the Cluster state using `kubectl get deployments,statefulsets,services,cm -o yaml`, for both group size cases, and for before and after this change. Everything was identical in the "after" case except for the expected:

- Removal of ports from headless server
- Change of `accessMode` for block devices to `ReadWriteOnce`
- Unique metadata

All volumes, PVCs, Pods, etc. were identical.

## Checklist

-----
[View rendered charts/cortx/README.md](https://github.com/keithpine/cortx-k8s/blob/CORTX-32157_data-refactor/charts/cortx/README.md)

### DIFF
--- a/charts/cortx/README.md
+++ b/charts/cortx/README.md
@@ -85,25 +85,21 @@ helm uninstall cortx
 | control.service.nodePorts.https | string | `""` | Node port for HTTPS for LoadBalancer and NodePort service types |
 | control.service.ports.https | int | `8081` | Control API service HTTPS port |
 | control.service.type | string | `"ClusterIP"` | Kubernetes service type |
-| cortxdata.confd.resources.limits.cpu | string | `"500m"` |  |
-| cortxdata.confd.resources.limits.memory | string | `"512Mi"` |  |
-| cortxdata.confd.resources.requests.cpu | string | `"250m"` |  |
-| cortxdata.confd.resources.requests.memory | string | `"128Mi"` |  |
-| cortxdata.cvgs | list | `[]` |  |
-| cortxdata.image | string | `"ghcr.io/seagate/centos:7"` |  |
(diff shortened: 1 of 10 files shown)


walterlopatka (approved):
lgtm

osowski (approved):
lgtm

PR #331, opened 2022-07-18
CORTX-33050: Add support for Motr log devices

keithpine:
## Description

This change updates the solution.yaml storage set definition to support "log" device entries, which are described in the latest [cluster.yaml sample file](https://github.com/Seagate/cortx-prvsnr/blob/e7677415f5d53003ee4baba9a7632eedc00415f0/conf/solution/cluster.yaml.sample#L22-L29). The format of the solution file has changed, see the Breaking change section. Log devices are specified as a list of devices, the same as data devices.

For now, this device type is optional. Omitting or including the log devices are both supported. The solution validation will not flag missing values.

I have **not** included a "name" field with the devices, as suggested. We can add this later in the future, but there is currently no use for it, so there is nothing for the deployment script or Chart to use it for, thus it should not be a required field.

With images 859 (planned v0.9.0 version) and the latest 867, I can see that the cluster.conf correctly sets the log entries. I don't believe Motr uses it for anything though. I am unable to verify that.

## Breaking change

The devices format of the storage sets has changed. Note that 0.9.0 already has a breaking change with respect to the formatting, so this can be included with that.

The `metadata` object used to take a singular device, now it is a list of devices. **Note:** motr setup will fail if more than one device is specified.

The `device` field of the device object is now more aptly named `path`.

## Type of change

- [x] Breaking change (bug fix or new feature that breaks existing functionality)

## Applicable issues

- This change fixes an issue: CORTX-33050
- This change is related to an issue: CORTX-32810

There aren't any versions that support log devices yet. The current version, 859, works with or without them.

## How was this tested?

I deployed in multiple ways:

- Image 859 with no log device
- Image 859 with a log device
- Image 867 with no log device
- Image 867 with a log device

In all cases, the behavior was the same: the cluster deployed successfully and scripts, S3, etc. were working. When log devices are specified, the statefulsets correctly mount all of the volume devices (meta + log + data), and the cluster.conf file is generated with an entry.

solution.yaml

```yaml
      devices:
        metadata:
        - path: /dev/sdc
          size: 25Gi
        log:
        - path: /dev/sdd
          size: 25Gi
        data:
        - path: /dev/sde
          size: 25Gi
```

cluster.yaml

```yaml
    storage:
    - name: cvg-1
      type: ios
      devices:
        metadata:
        - /dev/sdc
        log:
        - /dev/sdd
        data:
        - /dev/sde
```

cluster.conf

```yaml
    cvg:
    - devices:
        data:
        - /dev/sde
        log:
        - /dev/sdd
        metadata:
        - /dev/sdc
        num_data: 1
        num_log: 1
        num_metadata: 1
      name: cvg-1
      type: ios
```

The status script passes, as it verifies the number of PV and PVCs are created according to the solution file.

## Additional information

I have not added the log device entry to the README.md file or chart values.yaml file yet. It should be added when the feature is ready to be used.

Fix list formatting in cluster.yaml to match the sample and be more consistent.

Remove the `common.release` key from config.yaml since this matches the sample.

## Checklist


keithpine:
> The `metadata` object used to take a singular device, now it is a list of devices. **Note:** motr setup will fail if more than one device is specified.

Not sure if we should do anything about this? The list format was suggested in the Jira ticket, and it also makes the handling of the devices more consistent. Should I remove more than the 1st from the deployment script? Ignore all but the first in the Chart? Just document it in the README?

tanujashinde0405 (approved):

walterlopatka (approved):
lgtm

PR #333, opened 2022-07-19
CORTX-33054: Use Consul as Confstore for CORTX components

keithpine:
## Description

Configure all CORTX services to use the Consul Confstore URL. The components will read/write their configuration to/from Consul, instead of a local `/etc/cortx/cluster.conf` file.

## Breaking change

Requires new image versions, see solution.example.yaml or AppVersion for required image.

## Type of change

## Applicable issues

This change requires the following images:

- `cortx-data:902`
- `cortx-rgw:902`
- `cortx-control:902`

## How was this tested?

Deployed cluster successfully. Examined Consul KV and see conf being store there. Examine local persistent volume, no `cluster.conf` file.

Performed S3 IO and CSM API requests with success.

Used log collection script.

## Checklist

- [X] The `image` fields in [solution.example.yaml](../k8_cortx_cloud/solution.example.yaml) have been updated to use the required versions.
- [X] The `appVersion` field of the [Helm chart](../charts/cortx/Chart.yaml) has been updated to use the new CORTX version.


charts/cortx/templates/_images.tpl
```diff
@@ -67,7 +67,7 @@ Return the CORTX setup initContainer
   {{- if eq $image "ghcr.io/seagate/centos:7" }}
     - sleep $(shuf -i 5-10 -n 1)s
   {{- else }}
-    - /opt/seagate/cortx/provisioner/bin/cortx_deploy -f /etc/cortx/solution -c yaml:///etc/cortx/cluster.conf
+    - /opt/seagate/cortx/provisioner/bin/cortx_deploy -f /etc/cortx/solution -c $CONFSTORE_URL
```
osowski, line 70: Did you want to expand out the long names for the config switches here (this one may have slipped through since you have it on the rest of them)?

keithpine: I sure do.

keithpine: Oh, just remembered, `cortx_deploy` does not support long parameters. 😢

charts/cortx/templates/_helpers.tpl
```diff
@@ -269,3 +269,15 @@ Return the Motr-client endpoint port
...
+{{/*
+Create a URL CORTX Confstore URL
+*/}}
+{{- define "cortx.confstore.url" -}}
+{{- if .Values.consul.enabled -}}
+{{- printf "consul://%s-consul-server:8500/conf" (include "cortx.fullname" .) -}}
+{{- else -}}
+{{/* FIXME: how do we get the external service name? */}}
+{{- printf "yaml:///etc/cortx/cluster.conf" -}}
+{{- end -}}
```
keithpine, line 282: I'm not sure what to do here. We've never validated external Consul configurations, so this works as long as the CORTX images support backwards compatibility.

The consul endpoints provided by the user are just a list of tcp and http endpoints. The URL in line 278 is neither of those, but it corresponds to the http endpoint, so presumably we could try to search for it. https://github.com/Seagate/cortx-prvsnr/blob/a98e8cafee57e06a75626c31e02ddf4398724e3e/conf/solution/config.yaml.sample#L7

osowski: I think for now this is fine to tag it as-is with a future TODO. There aren't explicit requirements to support external Consul deployments or a Consul deployment that we (as the cortx-k8s deployer) don't control the service name of. So for the time being, what is written above is valid.

Once requirements change to support explicit external Consul deployments, then we can revisit that as a Helm Chart input and pulling that in directly via values.yaml.

k8_cortx_cloud/logs-cortx-cloud.sh
```diff
@@ -115,10 +115,20 @@ function tarPodLogs()
     local container=$1
 
     printf "\n ⭐ Generating support-bundle logs for pod: %s\n" "${pod}"
+    # shellcheck disable=SC2016
     kubectl exec "${pod}" -c "${container}" --namespace="${namespace}" -- \
-      cortx_support_bundle generate \
-        --cluster_conf_path yaml:///etc/cortx/cluster.conf \
-        --location file://${path} \
+      env location="file://${path}" \
```
keithpine, line 120: I don't really like this, so open to suggestions.

walterlopatka: I don't have any better ideas.  We'll probably want to take another look at logs-cortx-cloud.sh in the future anyway, so we might have an opportunity to change (or just remove the script).

walterlopatka (approved):

k8_cortx_cloud/logs-cortx-cloud.sh
osowski: If this works, I'm good with it for now (to avoid overthinking it). We'll want to reassess logs and support bundle as a whole anyway, as Walter mentions, so we can optimize it then.

osowski (approved):
lgtm

PR #339, opened 2022-07-21
CORTX-33842: Support multiple cortx-control pods

walterlopatka:
## Description
This change provides support for multiple cortx-control pods in a cluster. 
This is supported in helm value: `control.replicaCount`.
There is no support in solution.yaml for this at this time.

## Type of change

## Applicable issues

- This change fixes an issue: CORTX-33842

N/A

## How was this tested?
1. Deploy with 1 pod, 3 pods, 10 pods
2. Verify that all cortx-control pods are serving data
3. Verify that all cortx-control pods are in sync (all accept write requests, all return same GET responses)
4. Run basic k8s deploy regression test.

## Additional information
There may be issues with running multiple cortx-control pods in terms of the implementation of those containers.  I have done basic functional test, but I leave it to the team responsible for cortx-control implementation to do due diligence in developer test.

## Checklist

### DIFF
--- a/k8_cortx_cloud/start-cortx-cloud.sh
+++ b/k8_cortx_cloud/start-cortx-cloud.sh
@@ -28,7 +28,7 @@ if [[ ${deployment_type} != "data-only" ]]; then
     printf "########################################################\n"
     printf "# Start CORTX Control                                   \n"
     printf "########################################################\n"
-    expected_count=1
+    expected_count=$(helm get values cortx | yq .control.replicaCount)
     kubectl scale deploy cortx-control --replicas ${expected_count} --namespace="${namespace}"
 
     printf "\nWait for CORTX Control to be ready"
--- a/k8_cortx_cloud/start-cortx-cloud.sh
+++ b/k8_cortx_cloud/start-cortx-cloud.sh
@@ -28,7 +28,7 @@ if [[ ${deployment_type} != "data-only" ]]; then
     printf "########################################################\n"
     printf "# Start CORTX Control                                   \n"
     printf "########################################################\n"
-    expected_count=$(helm get values cortx | yq .control.replicaCount)
+    expected_count=$(helm get values cortx -n "${namespace}" | yq .control.replicaCount)
     kubectl scale deploy cortx-control --replicas ${expected_count} --namespace="${namespace}"
 
     printf "\nWait for CORTX Control to be ready"
--- a/k8_cortx_cloud/start-cortx-cloud.sh
+++ b/k8_cortx_cloud/start-cortx-cloud.sh
@@ -28,7 +28,7 @@ if [[ ${deployment_type} != "data-only" ]]; then
     printf "########################################################\n"
     printf "# Start CORTX Control                                   \n"
     printf "########################################################\n"
-    expected_count=$(helm get values cortx -n "${namespace}" | yq .control.replicaCount)
+    expected_count=$(helm get values cortx --all --namespace "${namespace}" --output yaml | yq .control.replicaCount)
     kubectl scale deploy cortx-control --replicas "${expected_count}" --namespace="${namespace}"
 
     printf "\nWait for CORTX Control to be ready"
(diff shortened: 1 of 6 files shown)


k8_cortx_cloud/start-cortx-cloud.sh
```diff
@@ -28,8 +28,8 @@ if [[ ${deployment_type} != "data-only" ]]; then
     printf "########################################################\n"
     printf "# Start CORTX Control                                   \n"
     printf "########################################################\n"
-    expected_count=1
-    kubectl scale deploy cortx-control --replicas ${expected_count} --namespace="${namespace}"
+    expected_count=$(helm get values cortx -n "${namespace}" | yq .control.replicaCount)
```
keithpine, line 31:
```suggestion
    expected_count=$(helm get values cortx --all --namespace "${namespace}" --output yaml | yq .control.replicaCount)
```

This doesn't work if you install with the default value. It will return "null" in that case. You need to use `--all` to show all computed values, which includes non-user values. Without it only shows the user values.

The default output is also "table" format, which seems to look like YAML, but it has some non-YAML output. `yq` is pretty tolerable with respect to its input, but it'd be safer to explicitly output as yaml.

k8_cortx_cloud/status-cortx-cloud.sh
```diff
@@ -114,6 +114,7 @@ else
 fi
 
 # Check pods
+expected_count=$(helm get values cortx -n "${namespace}" | yq .control.replicaCount)
```
keithpine, line 117: Same comment as above.

test/regression/test_deploy.py
```diff
@@ -10,12 +10,17 @@
 from utils import Logger, StopWatch
 
 
+def get_expected_cortx_control_pods(namespace):
+    cmd = f'helm get values cortx -n {namespace} | yq .control.replicaCount'
```
keithpine, line 14: Same comment as above. You could also go native Python and use JSON (or YAML).

```python
>>> with subprocess.Popen(["helm", "get", "values", "cortx", "--output", "json"], stdout=subprocess.PIPE) as proc:
...   print(json.loads(proc.stdout.read())["control"]["replicaCount"])
...
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
KeyError: 'replicaCount'
>>> with subprocess.Popen(["helm", "get", "values", "cortx", "--all", "--output", "json"], stdout=subprocess.PIPE) as proc:
...   print(json.loads(proc.stdout.read())["control"]["replicaCount"])
...
1
```

keithpine (changes requested):

k8_cortx_cloud/start-cortx-cloud.sh
walterlopatka: Good catch.  I only tested with non-default namespace when I made this last-minute change.  I'll fix it.

walterlopatka: I fixed it in the most recent commit.

test/regression/test_deploy.py
walterlopatka: In the end I kept this the same (but fixing the problem you point out).  Something wasn't quite working (between getting json from stdout and indexing into it), and the test cycle was kind of painful.  I went with the simpler working solution.

k8_cortx_cloud/status-cortx-cloud.sh
walterlopatka: Fixed in latest commit.

walterlopatka:
I have completed testing.  I think I have addressed all the comments and this should be ready to go.  @keithpine , thanks for the thorough review.

test/regression/test_deploy.py
```diff
@@ -10,12 +10,17 @@
 from utils import Logger, StopWatch
 
 
+def get_expected_cortx_control_pods(namespace):
+    cmd = f'helm get values cortx --all --namespace {namespace} --output yaml | yq .control.replicaCount'
+    child = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)  # nosec
+    return int(child.communicate()[0])
```
keithpine, line 16: FYI, this raises an Exception if the output is "null" for some reason. Probably a contrived example, but I ran this manually when the chart wasn't installed, which will reproduce it.

keithpine (approved):
I was able to deploy with a custom values file, see multiple replicas, scale them up and down, and successfully issue API requests. 👍

PR #358, opened 2022-08-22
Prototype:  K8s job to tar files in cortx PVC

walterlopatka:
## Description

CORTX-32645 requests a method for getting log files from CORTX
pods that are inaccessible (e.g. Terminated).  The files are
stored in PVs on the worker nodes.

The script has the idea of creating a job that starts a pod
that mounts the PVC that contains the files that the Admin
is after and tars the file to stdout.  Once the job completes
the script runs "kubectl logs" to get the tar file and write
it to local disk.

The CORTX PVCs are all simply named, so it is straightforward
for an admin to know and use the PVC name.  Examples:

cortx-ha  (for the cortx-ha pod)

data-cortx-server-[0-9] for cortx-server pods

data-cortx-data-g[0-9]-[0-9] for cortx-data pods

(cortx-control pods no longer use a backing PVC, so this is not included)

Really any PVC that can be mounted in the containers file system can be used.

Signed-off-by: Walter Lopatka <walter.lopatka@seagate.com>

### DIFF
--- a/k8_cortx_cloud/get-logs-from-pvc.sh
+++ b/k8_cortx_cloud/get-logs-from-pvc.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+
+pvc=$1
+namespace="${2:-default}"
+
+function help()
+{
+    echo "$0 <pvc> [<namespace>]"
+    exit 1
+}
+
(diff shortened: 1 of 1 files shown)


walterlopatka:
(Closed, as this was a prototype.)

PR #361, opened 2022-08-27
CORTX-32645: New script: get-logs-from-pvc.sh

walterlopatka:
## Description
This is a new tool to collect log files from PVC for the case when a Pod is not running and support bundle cannot collect
this data otherwise.  This is intended to be a support or advanced tool for a special circumstance.

This uses a Kubernetes job to start a Pod that mounts the PVC specified by the user.  It tars the contents of the PVC into tmp.  The script then kubectl cp's the tar file to the local host.

This script provides basic error checking, but there are some inputs that can cause problems, such as specifying a PVC that does not exist in the specified namespace.  In this case the script with hang with the collector pod in "Pending" state.  (This is pretty obvious from script stdout.)

## Type of change

## Applicable issues

- This change fixes an issue: CORTX-32645

N/A

## How was this tested?
Local testing of various cortx PVCs.

## Checklist


walterlopatka:
I'm not sure of this needs any additional documentation.  Maybe I should add something to the troubleshooting section of README.  Hopefully this tack of using a job to tar the PVC contents, and then kubectl cp to the local host is good enough to satisfy the request.

k8_cortx_cloud/get-logs-from-pvc.sh
```diff
@@ -0,0 +1,112 @@
...
+tar_complete="false"
+while [[ "${tar_complete}" == "false" ]]; do
+  sleep 1
+  kubectl exec "${pod_name}" -- ls /tmp/tarfile_created &> /dev/null && tar_complete=true
+done
+
+# Get logs, save to .tgz
+echo "Copying PVC contents to ${tarfile}"
+kubectl cp "${pod_name}":tmp/"${tarfile}" "${tarfile}"
+kubectl exec "${pod_name}" -- touch /tmp/stopme
```
osowski, line 109: Do you want to remove the tar file either here or after line 71 above, in order to prevent the potential buildup of tar files after subsequent runs?

osowski: Ah, you're right. I read the command too quickly and thought you were tar'ing inside the PVC itself. Based on being in `/tmp`, it's fine the way it is.

keithpine (changes requested):

k8_cortx_cloud/get-logs-from-pvc.sh
```diff
@@ -0,0 +1,112 @@
...
+job_name="cortx-log-${pvc}-${datestr}"
+tarfile="${pvc}.tgz"
+
+if [[ -f "${tarfile}" && "${force_overwrite}" == "false" ]]; then
+  printf "%s already exists. -f to overwrite.\n" "${tarfile}"
+  exit 1
+fi
+
+
+cat << EOF | kubectl apply -f - || true
```
keithpine, line 54: If the `kube apply` fails, should it really be ignored? I'd think the rest of the script would not function well if the job isn't created?

k8_cortx_cloud/get-logs-from-pvc.sh
```diff
@@ -0,0 +1,112 @@
...
+        - name: data
+          persistentVolumeClaim:
+            claimName: "${pvc}"
+      restartPolicy: OnFailure
+EOF
+
+
+# Get pod name
+pod_status="Starting"
+while [[ "${pod_status}" != "Running" ]]; do
```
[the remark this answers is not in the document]
keithpine: You'd still need to get the pod name, but I think you could just skip the manual loop and checking status otherwise, and it would also put a time limit on the wait, which the manual loop doesn't have.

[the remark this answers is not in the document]
keithpine: I tested `--pod-running-timeout`, and it doesn't work very well.

I'd suggest `kubectl wait --for=condition=ready --selector=job-name="${job_name}" --namespace "${namespace}" pod` instead to replace this while loop.

k8_cortx_cloud/get-logs-from-pvc.sh
```diff
@@ -0,0 +1,143 @@
...
+  pod_name="${my_array[0]}"
+done < <(kubectl get pods --namespace "${namespace}" --selector=job-name="${job_name}" --no-headers)
+
+
+printf "Waiting for tar to complete.\n"
+kubectl exec --namespace "${namespace}" "${pod_name}" -- sh -c 'until [ -f /tmp/tarfile_created ]; do sleep 1; done' || exit_msg "Failed waiting for job to complete" 1
+
+# Get logs, save to .tgz
+printf "Copying PVC contents to %s.\n" "${tarfile}"
+kubectl cp "${pod_name}":tmp/"${tarfile}" "${tarfile}"
```
keithpine, line 147:
```suggestion
kubectl cp "${namespace}/${pod_name}":tmp/"${tarfile}" "${tarfile}"
```

osowski (approved):
lgtm

osowski:
> lgtm

Well it looks good once the Shellcheck error is fixed.

k8_cortx_cloud/get-logs-from-pvc.sh
walterlopatka: I agree.  I tried to address it, but ran into a shellcheck obstacle.  I think this will have to remain as it is for now.

keithpine (approved):

PR #377, opened 2022-09-09
CORTX-34189 Fix status-cortx-cloud.sh regression (data-only deploy)

walterlopatka:
## Description
With v0.10.0 and the introduction of potentially multiple replicas of cortx-control pods, I introduced a regression in status-cortx-cloud.sh, such that the script reported failures with a data-only deployment. This is because logic was added to account for potentially multiple replicas did not take into account a possible data-only deployment.

Along with the fix I have added a basic regression test that deploys with data-only deployment.

## Type of change

## Applicable issues
- This change fixes an issue: CORTX-34189
- This change is related to an issue: #

N/A

## How was this tested?
Added new regression test that deploys with data-only deployment.
I also ran the existing regression test to confirm that standard deployment works as expected.

## Checklist

- [ x All commits are signed off and are in agreement with the [CORTX Community DCO and CLA policy](https://github.com/Seagate/cortx/blob/main/doc/dco_cla.md).

### DIFF
--- a/k8_cortx_cloud/status-cortx-cloud.sh
+++ b/k8_cortx_cloud/status-cortx-cloud.sh
@@ -115,6 +115,7 @@ fi
 
 # Check pods
 expected_count=$(helm get values cortx --all --namespace "${namespace}" --output yaml | yq .control.replicaCount)
+[[ ${data_deployment} == true ]] && expected_count=0
 count=0
 msg_info "| Checking Pods |"
 while IFS= read -r line; do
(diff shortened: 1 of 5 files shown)


k8_cortx_cloud/status-cortx-cloud.sh
```diff
@@ -115,6 +115,7 @@ fi
 
 # Check pods
 expected_count=$(helm get values cortx --all --namespace "${namespace}" --output yaml | yq .control.replicaCount)
+[[ ${data_deployment} == true ]] && expected_count=0
```
keithpine, line 118: Another approach would be to check the`.control.enabled` value, but I suppose solution.yaml should take precedence for the scripts.

keithpine (approved):

PR #385, opened 2022-09-16
CORTX-32850: Add Server container probes

keithpine:
## Description

Add container probes for the Server RGW containers. Each probe is an `httpGet` to the `/swift/healthcheck` API which returns always returns status `200 OK` when responding.

For now, the probes are disabled by default and will be enabled by default in the future (except the startup probe).

The probes are not exposed in the solution file, instead they can be customized using a custom values.yaml file.

## Type of change

## Applicable issues

- This change fixes an issue: CORTX-32850

## How was this tested?

- Deployed with defaults, no probes defined
- Deploy with probes on, confirmed statefulset defines probes
- Manually caused probes to fail by attaching gdb to radosgw process to make it hang
- Defined custom probes and observed they were applied correctly
- Defined custom probes that failed intentionally, observed containers restarted, not ready, etc.

## Additional information

The probes are currently disabled for now so we can gather some experience as to what the right default settings are. Once we have a better sense of the correct default values, we can enable the liveness and readiness probes by default (leaving startup disabled, assuming it's not really needed).

The custom probes do not support templates, it would be nice if the did. Consider that for future updates.

## Checklist

-----
[View rendered charts/cortx/README.md](https://github.com/keithpine/cortx-k8s/blob/CORTX-32850_rgw-probes/charts/cortx/README.md)

### DIFF
--- a/charts/cortx/README.md
+++ b/charts/cortx/README.md
@@ -157,8 +157,29 @@ helm uninstall cortx
 | server.persistence.accessModes | list | `["ReadWriteOnce"]` | Persistent volume access modes |
 | server.persistence.size | string | `"1Gi"` | Persistent volume size |
 | server.replicaCount | int | `1` | Number of Server replicas |
+| server.rgw.customLivenessProbe | object | `{}` | Override the default RGW liveness probe with a custom one. |
+| server.rgw.customReadinessProbe | object | `{}` | Override the default RGW readiness probe with a custom one. |
+| server.rgw.customStartupProbe | object | `{}` | Override the default RGW startup probe with a custom one. |
+| server.rgw.livenessProbe.enabled | bool | `false` | Enable the RGW container liveness probe |
+| server.rgw.livenessProbe.failureThreshold | int | `5` | Number of times to retry the liveness probe after it fails, before the pod is marked Unready. |
+| server.rgw.livenessProbe.initialDelaySeconds | int | `10` | Number of seconds to wait before the liveness probe is initiated |
+| server.rgw.livenessProbe.periodSeconds | int | `5` | How often (in seconds) to perform the liveness probe |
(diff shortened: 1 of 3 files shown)


charts/cortx/README.md
```diff
@@ -157,8 +157,29 @@ helm uninstall cortx
...
 | server.persistence.size | string | `"1Gi"` | Persistent volume size |
 | server.replicaCount | int | `1` | Number of Server replicas |
+| server.rgw.customLivenessProbe | object | `{}` | Override the default RGW liveness probe with a custom one. |
+| server.rgw.customReadinessProbe | object | `{}` | Override the default RGW readiness probe with a custom one. |
+| server.rgw.customStartupProbe | object | `{}` | Override the default RGW startup probe with a custom one. |
+| server.rgw.livenessProbe.enabled | bool | `false` | Enable the RGW container liveness probe |
+| server.rgw.livenessProbe.failureThreshold | int | `5` | Number of times to retry the liveness probe after it fails, before the pod is marked Unready. |
+| server.rgw.livenessProbe.initialDelaySeconds | int | `10` | Number of seconds to wait before the liveness probe is initiated |
+| server.rgw.livenessProbe.periodSeconds | int | `5` | How often (in seconds) to perform the liveness probe |
+| server.rgw.livenessProbe.successThreshold | int | `1` | Number of times the liveness probe must succeed, after having failed, to be considered successful (must be 1) |
```
walterlopatka, line 167: Just curious... why "must be 1"?

osowski: Per the [spec](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes), this must be 1 for liveness and startup probes.

osowski (approved):

charts/cortx/README.md
keithpine: I considered removing it, but kept it for consistency. Actually, the template implementation should updated to use a macro, which is why it's left in.

walterlopatka (approved):

